PR #121, opened 2018-08-10
feat(build): Build for linux distros

mrmeku:


mrmeku:
Questions to answer: Is it worth targetting "deb" and "snap" or should we just target only "AppImage" instead? "AppImage" is ubiquitous but unfamiliar to many Ubuntu users. "snap" is incredibly popular, but not quite as capable as "AppImage".

PR #123, opened 2018-08-10
feat(ui): Alter UX of opening workspace in external program

mrmeku:
User feedback noted that it was hard to identify having just the external program's icon as a button. To compensate we now use a more standard button UI which opens a menu of available programs to open.


mrmeku:
image
image

apps/angular-console-e2e/src/integration/workspace.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { openProject, projectPath, uniqName } from './utils';
+
+describe('Workspaces', () => {
```
vsavkin: I think this is a duplicate of projects.spec.ts

PR #131, opened 2018-08-10
 feat(server): add server.dev mode with nodemon

beeman:
This PR a `server.dev` mode that allows for running the server with nodemon.


vsavkin:
Thank you for the PR.

We changed the setup to always use electron when running the server to be closer how the app actually runs in production. So I'm afraid using ts-node will no longer work.

beeman:
Ok, that makes sense @vsavkin. 

I will check the new setup :)

Issue #80, opened 2018-08-06
task-runner component: Task Output is off-centre


mrmeku:
Since this is an electron app we should try to align the terminal title with the app window title

PR #148, opened 2018-08-13
style:Task Output label in task-runner component is off-centre

stefanoslig:
A minor css fix in the task-runner component (#80 )


mrmeku:
Closing this PR in favor of when which guarantees that the title is absolutely centered within the window: https://github.com/nrwl/angular-console/pull/152

PR #158, opened 2018-08-15
Use APOLLO_OPTIONS

kamilkisiela:
Hi guys,

thanks for using Apollo! :) I refactored a bit the way you initialize Apollo. There's a InjectionToken for it so it all runs on DI level.

I've got a tons of new features for you and I'm going to push it one by one, starting soon.


vsavkin:
Great! Thank you.

PR #159, opened 2018-08-15
feat: add save/load of electron window size and location

### LINKED ISSUE #92: Save window size when exiting app
Feature request:

Would like the app to save the window dimensions when exiting the app and use those when opening the app if they exist. It would also be nice if it saved the window location as well.

jschwarty:
When closing the Electron app the window size and location is saved to the store and is used when opening the app to restore size and location.

Closes #92


Yonet:
LGTM, thanks!

PR #166, opened 2018-08-18
refactor:cancel polling 

stefanoslig:
This is just an enhancement. When you enter into a view where there is a poller that is running, the poller continues running after leaving this view.

### DIFF
--- a/libs/feature-extensions/src/lib/extensions/extensions.component.ts
+++ b/libs/feature-extensions/src/lib/extensions/extensions.component.ts
@@ -1,16 +1,17 @@
-import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { ChangeDetectionStrategy, Component, OnDestroy } from '@angular/core';
 import { ActivatedRoute, NavigationEnd, Router } from '@angular/router';
 import { Extension, ExtensionGroup, Schematic } from '@angular-console/schema';
 import { Apollo } from 'apollo-angular';
 import gql from 'graphql-tag';
-import { combineLatest, Observable } from 'rxjs';
+import { combineLatest, Observable, Subject } from 'rxjs';
 import {
   distinctUntilChanged,
   filter,
   map,
   shareReplay,
   startWith,
-  switchMap
+  switchMap,
+  takeUntil
 } from 'rxjs/operators';
 import { TaskCollection, TaskCollections, Task } from '@angular-console/ui';
 
@@ -24,7 +25,8 @@ interface ExtensionId {
   templateUrl: './extensions.component.html',
   styleUrls: ['./extensions.component.scss']
 })
-export class ExtensionsComponent {
+export class ExtensionsComponent implements OnDestroy {
+  destroy$ = new Subject<void>();
   private readonly extensions$: Observable<
     Array<ExtensionGroup>
   > = this.route.params.pipe(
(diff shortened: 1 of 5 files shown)


vsavkin:
Thank you for your PR!

Let's talk a bit about the issue you are fixing.

Some background:

On most pages, there are several graphql queries running. For instance, if I go to the generate screen, the app will have two queries running:

1. One fetching the workspace name
2. The other one fetching the list of schematics
But, if you go to the screen for an individual schematic, only you should see only one query running (the one fetching the workspace name).

We can have more queries running on the same page--that's OK. We don't want to have leaks, where an observable is never unsubscribed, and as the user navigates back and forth, the number of observables keeps going up.

To prevent that, we are trying to use the async pipe instead of calling `subscribe`. We don't use it everywhere, but we do use it in say SchematicComponent. This means that the pipe will unsubscribe as you navigate away from that page. So I don't think adding takeUntil is necessary in that component. I don't think it is leaking subscriptions. 

Does it make sense? Could you provide a repro of a subscription leak?

stefanoslig:
Hello Victor,

Thanks for your response. I had noticed that you try to unsubscribe with the async pipe. Actually, you are right, but the problem that I mentioned exists for the components I have in my PR except the "projects.component.ts". So, for the rest components, I will tell you which is the problem and I will give two ways to reproduce it.

In Schematics component you have the schematicCollections$ stream in line 29. Then in line 94 you have the taskCollections$ stream. Then you pass the latter(taskCollections$) as an Observable in the template of this component (<ui-task-selector [taskCollections$]="taskCollections$") and inside "task-selector.component.ts" you unsubscribe from taskCollections$ Observable. The problem is that there isn't someone that has unsubscribed from the schematicCollections$ stream. 

How to reproduce it:
Create/Import two workspaces(let's say test1 and test2). 
Now in "Recent " tab you have two workspaces. Enter into test1 and navigate to schematics. 
Then go back to Recent and enter into test2.
Open your dev tools and go to network. You will see the query for  schematicCollections appearing with two different paths(../../test1,  ../../test2)

A second way to reproduce it is to go to the SchematicComponent and to add take(1) in the end of schematicCollections query Navigate to schematics and then to a specific schematic . Even if you have added take(1) in the query, the poller continues to emit values. This is the query from the  SchematicsComponent.

We have the same behavior in ExtensionsComponent, TargetsComponent and WorkspaceComponent(I didn't have enough time to test WorkspaceComponent) 

There is also an open subscription in line 94 of NewWorkspaceComponent. The async pipe here is not enough.

I hope that I'm not wrong. I will fix ProjectsComponent which hasn't any problem.

vsavkin:
Thank you! I can reproduce the issue. But I'm not sure why this issue is happening.

- schematicCollections$ is an input into taskCollections$. 
- taskCollections$ uses shareReplay, which means that is should create a single subscription of schematicCollections.
- taskCollections$ is then subscribed once directly (with take(1)) and used to construct filteredTaskCollections$
- filteredTaskCollections$ is used with the async pipe

Give this setup, the only explicit subscription we have is inside ngOnInit of `TaskSelectorComponent`.  Because we use take(1), it should be fine. 

In general, since we don't subscribe `schematicCollections$`, we should not unsubscribe from it either. Whatever creates the subscription should release it. 

Maybe we don't use operator correctly and there is a subscription lost somewhere in one of them.

Makes sense?

stefanoslig:
Yes, I agree with you that we should not unsubscribe from something that does not have a subscription on it. To be honest, it isn't clear what causes the problem to me either. I'll give a try to find what's going on. You can close this PR in any case.

vsavkin:
@stefanoslig I'll close this PR. Please let me know if you made any progress in figuring out why the issue occurs. It's an tricky one.

PR #171, opened 2018-08-20
feat: use electron store to store settings instead of local storage

vsavkin:


vsavkin:
@Yonet you are seeing an existing issue. If you run `yarn start mac.start-electron` (or in general start console from the terminal), it will add the project you are starting it from to "recents". It does it with the favorite flag set to false.

I fixed the issue (so it no longer reads the same project to recents) and added a test for it as well.

CONTRIBUTING.md
```diff
@@ -15,7 +15,9 @@ After cloning the project run: `yarn`.
 
 After that, run `yarn start dev.prepare`. Every time you add or remove dependencies in electron/package.json, you will need to rerun `dev.prepare`. 
 
-After this, run `yarn start dev.up` to start the dev environment. The application will start the process listening on port 4200.
+After this, run `yarn start dev.up` to start the dev environment. The application will start the process listening on port 4200. 
+
+You can also run the electron app by running `yarn start mac.start-electron` or `yarn start win.start-electron`.
```
Yonet: Thank you!

PR #174, opened 2018-08-21
fix: split fields of "arguments" type

vsavkin:


libs/utils/src/lib/serializer.service.ts
```diff
@@ -61,24 +61,24 @@ export class Serializer {
...
+        args = [...args, ...value[f.name].split(' ').filter((r: any) => !!r)];
+      } else {
+        args.push(`--${f.name}=${value[f.name]}`);
+      }
+    });
     return args;
   }
 
   argsToString(args: string[]): string {
+    console.log(args);
```
mrmeku: Remove this line

PR #175, opened 2018-08-21
Apollo Server 2.0, latest Apollo Angular, refactoring, SDL

kamilkisiela:
- Apollo Server 2.0
- [Latest Apollo Angular](https://medium.com/the-guild/apollo-angular-code-generation-7903da1f8559)
- refactoring - moved files under /api directory
- used SDL instead of classes from `graphql`
- separated resolvers

## Why ApolloServer?

It's just easier to set up, I'm sure you want to implement subscription soon because of polling and it has it built in. And a lot of other things.

## Latest Apollo Angular

I released a new version today, it allows to use code generation which might improve your code by making it even more strongly typed and by generating a Query, Mutation and Subscription services you no longer have to use `Apollo` service, pass a document and add interfaces.

[You can see in details what I mean by that.](https://medium.com/the-guild/apollo-angular-code-generation-7903da1f8559)

## Why SDL?

It just looks good, allows to easily find what you're looking for in a schema, it allows to keep resolvers separated.

## Resolvers separated?

Thanks to tools like `graphql-code-generator` it's possible to generate interfaces for those resolvers so TypeScript gives you more confidence about the code.

---

I mentioned few helpful features and techniques here and if time allows and you're interested in it I will try to implement those soon.


vsavkin:
> I'm sure you want to implement subscription soon 
Yup. We definitely want that.

@kamilkisiela thank you so much, Kamil. I'll merge the PR.

PR #181, opened 2018-08-27
feat(build): Successfully build and install image on arch linux

mrmeku:


vsavkin:
Per our discussion:

* We should limit the linux distributions to only two packages
* I don't see the icon being provisioned in the linux configuration in package.json. Does the icon work?
* Search for "implement linux support" in the code base. All of them should be implemented.

package-scripts.js
```diff
@@ -55,7 +66,19 @@ module.exports = {
       'start-server': 'electron dist/electron --server',
       'start-electron': 'NODE_ENV=development electron dist/electron',
       'builder-dist': 'electron-builder --mac -p never',
-      'builder-dist-linux': 'electron-builder --linux -p never'
+    },
+    linux: {
```
vsavkin: Linux and mac support are more or less the same. Do you think it makes sense to create "unix" section and have linux and mac scripts delegate to "unix"?

mrmeku: I'll tackle this in a followup PR

PR #185, opened 2018-08-28
Use code generation to generate types for resolver and whole schema

kamilkisiela:
It makes resolvers strongly typed. You can also use it on the client side, to generate types for each GraphQL query and mutation.

Thanks to that I noticed that `readArchitect` doesn't return a `description` property, it's always undefined but it's defined in GraphQL Schema.

https://github.com/nrwl/angular-console/blob/2783d9b590b18b9b2263d539ece9700a6cd0ab1a/server/src/api/read-projects.ts#L17-L24


server/code-gen.ts
```diff
@@ -0,0 +1,20 @@
+import { generate } from 'graphql-code-generator';
```
vsavkin: Could you rename the file into something like `gen-graphql-types`?

server/src/api/read-projects.ts
```diff
@@ -40,6 +20,7 @@ function readArchitect(project: string, basedir: string, architect: any) {
       ? Object.keys(value.configurations).map(name => ({ name }))
       : [];
     return {
+      description: '',
```
vsavkin: The issue with the description field is a good catch. But I think the proper fix is to change the schema to remove the description field. It's actually not available in angular.json

vsavkin:
@kamilkisiela any update on this? It's a good change, so I'd like to merge it before the next release. If it's OK by you, I can also make the change myself and merge it.

kamilkisiela:
@vsavkin Hi Victor, I forgot about it but I'm going to make changes today

server/src/schema/index.ts
```diff
@@ -3,6 +3,6 @@ import { resolvers } from './resolvers';
 import { typeDefs } from './schema';
 
 export const schema = makeExecutableSchema({
-  resolvers,
+  resolvers: resolvers as any,
```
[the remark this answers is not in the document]
kamilkisiela: That's because generated objects has no index signature.

```typescript
interface Resolvers {
  [key: string]: any // <-- that's missing but that's fine since it would allow to add non existing keys
}
```

server/src/api/read-projects.ts
kamilkisiela: I also removed resolver for `description` field.

vsavkin:
Merged here: https://github.com/nrwl/angular-console/commit/65ac7bf2aa8f907286d016c430dca5fafeeaf099

Thank you, @kamilkisiela .

PR #191, opened 2018-09-06
chore(build): create docker-based linux build process

jeffbcross:


package-scripts.js
```diff
@@ -18,7 +18,7 @@ function withPlatform(command) {
 }
 
 function electronBuilder(dashP) {
```
vsavkin: could you rename dashP into publich?

mrmeku:
@jeffbcross Is this PR still relevant or has our build process changed enough that we need to start over?

PR #216, opened 2018-09-22
fix: not all schematic collections get picked up and displayed

vsavkin:


server/src/api/read-schematic-collections.ts
```diff
@@ -22,10 +27,30 @@ interface Schematic {
...
-export function readSchematicCollections(
+export function readAllSchematicCollections(basedir: string) {
+  try {
+    const nodeModulesDir = path.join(basedir, 'node_modules');
+    const packages = listOfUnnestedNpmPackages(nodeModulesDir);
+    const schematicCollections = packages.filter(p => {
+      try {
+        return !!readJsonFile(path.join(p, 'package.json'), nodeModulesDir).json
+          .schematics;
+      } catch (e) {
```
mrmeku: Can you leave a comment saying what type of error we expect here? If the type of errors are unexpected maybe we should log them rather than silently filtering them out

server/src/api/read-schematic-collections.ts
```diff
@@ -22,10 +27,30 @@ interface Schematic {
...
+        return false;
+      }
+    });
+    return schematicCollections.map(c =>
+      readSchematicCollections(nodeModulesDir, c)
+    );
+  } catch (e) {
```
mrmeku: What's the point of this try catch if we just rethrow?

server/src/utils.ts
```diff
@@ -68,7 +70,23 @@ export function findClosestNg(dir: string): string {
   }
 }
 
-export function listFilesRec(dirName: string): string[] {
+export function listOfUnnestedNpmPackages(nodeModulesDir: string): string[] {
+  const res: string[] = [];
+  fs.readdirSync(nodeModulesDir).forEach(npmPackageOrScope => {
+    if (npmPackageOrScope.startsWith('@')) {
```
mrmeku: This seems a bit angular and nrwl specific in terms of the naming convention. I think doing a breadth-first search of node modules and stopping going deeper when you encounter a package.json is what you want here.

server/src/utils.ts
```diff
@@ -182,3 +236,21 @@ export function normalizePath(value: string): string {
...
     .join('\\');
 }
+
+/**
+ * To improve performance angularconsole preprocesses
+ *
+ * * the list of local files
+ * * json files from node_modules we are likely to read
+ *
+ * both the date sets get updated every 30 seconds.
```
mrmeku: data?

server/src/api/read-schematic-collections.ts
vsavkin: Added a check to ignore only the right type of error

server/src/api/read-schematic-collections.ts
vsavkin: Removed it.

server/src/utils.ts
vsavkin: This is actually not an angular convention. This is how node_modules are structured.

All top-level npm packages are either direct children of `node_modules` or located in a folder starting with a @, which is a direct child of `node_modules`.

server/src/utils.ts
vsavkin: Fixed it

apps/angular-console-e2e/src/integration/extensions.spec.ts
```diff
@@ -33,24 +34,28 @@ describe('Extensions', () => {
...
-    );
+    clickOnTask('Available Extensions', '@angular/material', false);
+    cy.get('div.context-title').contains('@angular/material');
 
     cy.get('button')
       .contains('Add')
       .click();
 
-    cy.wait(100);
+    cy.wait(120000);
```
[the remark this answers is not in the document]
vsavkin: Fixed it in a nice way. Now we just wait for the command to complete. I also updated another test to use the same utility.

PR #217, opened 2018-09-22
fix: scripts section gets removed from package.json

### LINKED ISSUE #203: scripts section is missing from package.json
Compare project created by Angular Console and Angular CLI. I found that scripts section is missing if the project is created by Angular Console. It seems the `@schematics/angular` that include in Angular Console is different than Github version.

image

vsavkin:
Closes #203

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -86,7 +86,7 @@ module.exports = {
       'clean': 'if exist dist rmdir dist /s /q',
       'compile': 'tsc -p server/tsconfig.json',
       'copy-assets': 'copy server\\package.json dist\\server\\package.json && (robocopy server\\assets dist\\server\\assets /e || echo 0)',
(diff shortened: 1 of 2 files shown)


package-scripts.js
```diff
@@ -86,7 +86,7 @@ module.exports = {
       'clean': 'if exist dist rmdir dist /s /q',
       'compile': 'tsc -p server/tsconfig.json',
       'copy-assets': 'copy server\\package.json dist\\server\\package.json && (robocopy server\\assets dist\\server\\assets /e || echo 0)',
-      'install-node-modules': 'cd dist\\electron && yarn',
+      'install-node-modules': 'cd dist\\server && yarn',
```
mrmeku: Just making sure, was this intentional? It seems unrelated

vsavkin: Yup. It's intentional.

PR #224, opened 2018-10-02
feat(ui): Add App Shell

vsavkin:


vsavkin:
We will be a bit more conservative and will start with: https://github.com/nrwl/angular-console/pull/232.

I'll reopen the PR once I do more measurements.

PR #227, opened 2018-10-03
build: add start-dev-server script that enables inspect for electron server

jaysoo:
Added `start-dev-server` script, as there isn't a good way (that I know of) to chain pass command line arguments from `nps`.


vsavkin:
can we change `start-server` to always start with `--inspect`? I don't think it is used anywhere except dev.server

jaysoo:
Oh ok, yeah I can do that.

jaysoo:
@vsavkin Done, updated `start-server`.

PR #230, opened 2018-10-03
fix: guard against schematics that do not have schema defined

jaysoo:
The schema entry for defining options may not be present, so it's best to skip over these entries.

This fixes the issue where Generate Code breaks after `@angular/elements` schematics are installed. Not sure if there is a way we can still display schematics from elements without having the schema.


vsavkin:
Fixed here: 20324c37bdb8178fc7fa0f3e31514cdd3e78668c

PR #241, opened 2018-10-05
feat(node): Add api endpoints to detect/install node

mrmeku:


apps/angular-console/src/app/app.module.ts
```diff
@@ -88,7 +89,12 @@ export function initApollo(
         {
           path: '',
           children: workspaceRoutes,
-          canActivateChild: [CancelCommandGuard]
+          canActivateChild: [IsNodeJsInstalledGuard, CancelCommandGuard]
+        },
+        {
+          path: 'install-nodejs',
+          loadChildren:
+            '@angular-console/feature-install-node-js#FeatureInstallNodeJsModule'
```
vsavkin: This is funky looking URL!!

vsavkin: Actually, it is pretty good

mrmeku: :p, don't blame me, blame the author of @angular/router

apps/angular-console/src/app/app.module.ts
```diff
@@ -88,7 +89,12 @@ export function initApollo(
...
         },
         {
           path: 'support',
```
vsavkin: We should probably make support lazy as well in a later PR

libs/feature-install-node-js/src/lib/feature-install-node-js.module.spec.ts
```diff
@@ -0,0 +1,14 @@
+import { async, TestBed } from '@angular/core/testing';
```
vsavkin: Delete an empty test

libs/feature-install-node-js/src/lib/install-node-js.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
```
vsavkin: Delete an empty spec

libs/feature-install-node-js/src/lib/install-node-js.component.ts
```diff
@@ -0,0 +1,71 @@
+import { Component } from '@angular/core';
+import { Router } from '@angular/router';
+import { Apollo } from 'apollo-angular';
+import gql from 'graphql-tag';
+import { Observable } from 'rxjs';
+import { map, switchMap, tap } from 'rxjs/operators';
+
+import { InstallNodeJsStatus } from '../../../../server/src/graphql-types';
```
vsavkin: could you remove this and replace this with any for now. I don't think the client-side code should reference the server folder.

Once nx 6.4 is out, I'm going to upgrade console to use it, so we can share libs between the server and the client. Then we can share all the types.

libs/feature-install-node-js/src/lib/install-node-js.component.ts
```diff
@@ -0,0 +1,71 @@
...
+                }
+              `
+            })
+            .valueChanges.pipe(
+              map(
+                (result: {
+                  data: { installNodeJsStatus: InstallNodeJsStatus };
+                }) => result.data.installNodeJsStatus
+              ),
+              tap(res => console.log(res))
```
vsavkin: this should not be here

libs/feature-install-node-js/src/lib/install-node-js.component.ts
```diff
@@ -0,0 +1,71 @@
...
+            ) as Observable<InstallNodeJsStatus>;
+        }),
+        tap(({ success }: InstallNodeJsStatus) => {
+          if (success) {
```
vsavkin: I think we should handle error as well.

You can still redirect to console but you can also use messanger service to show some error message.

libs/utils/src/lib/nodejs-install.service.ts
```diff
@@ -0,0 +1,37 @@
+import { Injectable } from '@angular/core';
+import gql from 'graphql-tag';
+
+import { CommandRunner } from './command-runner.service';
+
+@Injectable({ providedIn: 'root' })
+export class NodejsInstall {
```
vsavkin: This service isn't used.

server/src/api/install-nodejs.ts
```diff
@@ -0,0 +1,82 @@
...
+  nodeDownload = null;
+  nodeInstallDone = false;
+  try {
+    switch (platform()) {
+      case 'darwin':
+        nodeDownloadProgress = progress({
+          length: 15546624,
+          time: 100
+        }) as Progress;
+        const pkg = fs.createWriteStream('node.pkg').on('finish', () => {
```
vsavkin: you should use the tmp directory. you don't want to write it into the current directory

vsavkin: You can use `import { dirSync } from 'tmp';`

apps/angular-console/src/app/app.module.ts
vsavkin: :) I was just testing out hte new vscode github integration

https://code.visualstudio.com/blogs/2018/09/10/introducing-github-pullrequests

Turns out you can add comments, but you cannot delete them :)

vsavkin:
Delete node.pkg file

vsavkin:
Looks very good. Great job, Dan! I left a few minor comments.

vsavkin:
Just tried it on Windows and I'm seeing the following issue:

The frontend keeps receiving these:

```
{downloadPercentage: 100, downloadSpeed: 114403.06086956522, success: null, cancelled: null, error: null, …}
```

I imagine is because this line `exec(`msiexec /i "${msi.path}"`);` needs to listen to `on('close')` to set nodeInstallDone to true.

apps/angular-console/src/app/app.module.ts
mrmeku: Agreed, should speed up boot up time a bit more (not that Jason hasn't already solved that)

libs/feature-install-node-js/src/lib/install-node-js.component.ts
mrmeku: Donezo

libs/feature-install-node-js/src/lib/install-node-js.component.ts
mrmeku: donzo

libs/utils/src/lib/nodejs-install.service.ts
mrmeku: Donezo

vsavkin:
Merged here:
https://github.com/nrwl/angular-console/pull/242

Made two changes:
- Fixed the window version (msi.path is absolute)
- I also added an option to skip the install. This is in case they don't have node in path, which is not a completely unthinkable thing.

Cheers!

PR #251, opened 2018-10-10
WIP: feat(tasks): Add UI for background tasks

mrmeku:
Pending some edits to our backend, this should more or less work once we have the graphql endpoints in place


libs/feature-action-bar/src/lib/action-bar.component.ts
```diff
@@ -0,0 +1,162 @@
...
+      ),
+      transition(`* <=> *`, animate(`250ms ease-in-out`))
+    ])
+  ]
+})
+export class ActionBarComponent {
+  @HostBinding(`@${ACTION_BAR_ANIMATION_NAME}`)
+  actionPanelAnimation: ActionBarAnimationStates = 'hidden';
+
+  CommandStatus = CommandStatus;
```
vsavkin: you will need to use a different enum here as the backend returns slightly different values

libs/feature-action-bar/src/lib/action-bar.component.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  private readonly actionsPanelShowing = new BehaviorSubject(false);
+  readonly actionsPanelShowing$ = this.actionsPanelShowing.asObservable().pipe(
```
vsavkin: I don't think you need asObservable, right?

libs/feature-action-bar/src/lib/action-bar.component.ts
```diff
@@ -0,0 +1,162 @@
...
+      case '0':
+        return '#D63965';
+      case '1':
+        return '#52B9D0';
+      case '2':
+        return '#6040B0';
+    }
+  }
+
+  getNameForWorkspace(path: string) {
```
vsavkin: I imagine this is temporary :)

libs/feature-extensions/src/lib/extension/extension.component.ts
```diff
@@ -153,6 +153,6 @@ export class ExtensionComponent implements OnInit {
   }
 
   onStop() {
-    this.runner.stopCommand();
+    this.runner.stopCommand('');
```
vsavkin: I removed all onStop

libs/ui/src/lib/terminal/terminal.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="command">$ {{ command }}</div>
+<div *ngIf="command" class="command">$ {{ command }}</div>
```
vsavkin: why would not command be defined?

libs/utils/src/lib/command-runner.service.ts
```diff
@@ -6,28 +6,40 @@ import { BehaviorSubject, interval, Observable, of } from 'rxjs';
 import { concatMap, map, takeWhile } from 'rxjs/operators';
 
 export interface CommandOutput {
-  status: 'success' | 'failure' | 'inprogress';
+  status: 'terminated' | 'success' | 'failure' | 'inprogress';
   out: string;
 }
 
+export enum CommandStatus {
```
vsavkin: This will need to change based on the PR I merged

libs/utils/src/lib/command-runner.service.ts
```diff
@@ -6,28 +6,40 @@ import { BehaviorSubject, interval, Observable, of } from 'rxjs';
...
+  STOPPED = 'stopped',
+  SUCCESSFUL = 'successful',
+  FAILED = 'failed',
+  IN_PROGRESS = 'running'
+}
+
+export interface Command {
```
vsavkin: This too

libs/ui/src/lib/terminal/terminal.component.ts
```diff
@@ -40,15 +40,21 @@ export class TerminalComponent implements AfterViewInit, OnDestroy {
   @Input() command: string;
 
   @Input()
-  set input(s: string) {
+  set inputStream(s: string) {
```
vsavkin: I know what you are trying to indicate here, but I think the stream suffix can be misleading. Folks often refer to Observables as streams. That's why I thought when I looked it it. 

We use the suffix chunk on the backend. Maybe it's better?

We can also change it around and actually pass an observable here and subscribe to it here. Currently we use |async pipe everywhere to get the input from the observable.

libs/feature-action-bar/src/lib/action-bar.component.ts
mrmeku: Fixed!

libs/ui/src/lib/terminal/terminal.component.html
mrmeku: It is not defined in the background task UI since it is displayed in the list of tasks itself

libs/ui/src/lib/terminal/terminal.component.ts
mrmeku: I changed to out / outChunk so that the naming conventions used on the server and frontend are the same

apps/angular-console-e2e/src/integration/utils.ts
```diff
@@ -78,17 +94,17 @@ export function projectNames(callback: (s: any) => void) {
...
 
 export function goToExtensions() {
-  cy.get('button#go-to-extensions').click();
+  cy.get('button#go-to-extensions').click({ force: true });
   waitForAnimation();
 }
 
 export function goToTasks() {
-  cy.get('button#go-to-tasks').click();
+  cy.get('button#go-to-tasks', { timeout: 3000 }).click({ force: true });
```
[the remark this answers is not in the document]
mrmeku: That should have been deleted

mrmeku:
Before submitting:

Need to implement a delete all recent tasks graphql endpoint and hook up to button

Need to debug why terminal output displaying is flaky

PR #255, opened 2018-10-11
feat: implement concurrent commands

vsavkin:


server/src/api/commands.ts
```diff
@@ -1,49 +1,117 @@
...
+    c.commandRunning = c.factory();
+  }
+
+  addOut(id: string, out: string) {
+    const c = this.findMatchingCommand(id, this.commandInfos);
+    c.out += out;
+    c.outChunk += out;
+  }
+
+  setFinalStatus(id: string, status: string) {
```
mrmeku: It feels like it would be nice to use an enum type here which the front end could use too. Could you maybe add a TODO here so we remember to clean that up later?

server/src/api/commands.ts
```diff
@@ -1,49 +1,117 @@
...
+      command,
+      status: 'waiting',
+      out: '',
+      outChunk: '',
+      factory,
+      commandRunning: null
+    });
+  }
+
+  restartCommand(id: string) {
```
mrmeku: When you restart a command I would expect that its output be cleared. Can you add

c.out = '' ?

server/src/api/commands.ts
```diff
@@ -1,49 +1,117 @@
 import * as os from 'os';
 
-const spawn = require('node-pty-prebuilt').spawn;
-
-interface CommandResult {
+interface CommandInformation {
+  id: string;
+  workspace: string;
```
mrmeku: This is the full workspace path correct? (Might be good to change the name to workspacePath to differentiate between workspace names)

libs/utils/src/lib/command-runner.service.ts
```diff
@@ -65,17 +78,90 @@ export class CommandRunner {
...
+    return this.apollo
+      .watchQuery({
+        query: gql`
+          query($id: String) {
+            commands(id: $id) {
+              id
+              workspace
+              command
+              status
+              out
```
mrmeku: If performance becomes an issue we should make it such that we only pull to get the full output on the first emission. Can you leave a comment here to that effect?

server/src/api/commands.ts
vsavkin: this is a workspace name (from package.json), we can replace it with 'path' if it helps

libs/utils/src/lib/command-runner.service.ts
vsavkin: I'll leave a comment.

PR #256, opened 2018-10-11
feat: add detailed status for commands

jaysoo:
- [x] build status
- [x] server status
- [x] test status


jaysoo:
Need to rebase and change the UI a bit with Dan's latest changes.

libs/utils/src/lib/feature-flags.service.ts
```diff
@@ -0,0 +1,3 @@
+export class FeatureFlags {
```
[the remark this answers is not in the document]
jaysoo: Rebased with master, this is gone now!

jaysoo:
Need to make sure tests pass and UI is intuitive to user.

.circleci/config.yml
```diff
@@ -13,10 +13,7 @@ jobs:
       - checkout
       - restore_cache:
           name: Restore Global Yarn Package Cache
-          keys:
```
mrmeku: I think you might need to rebase again? Or was this change deliberate?

libs/feature-action-bar/src/lib/action-bar.component.html
```diff
@@ -22,7 +22,7 @@
   <!-- Task list -->
   <mat-list [@growShrink]="(showActionList$ | async) ? 'expand' : 'contract'">
     <ng-container *ngFor="let action of actions; trackBy: trackByCommandId" [ngSwitch]="action.status">
-      <mat-list-item @growShrink matRipple #itemRipple="matRipple" (click)="itemRipple.disabled ? '' : toggleItemExpansion(action.id)">
+      <mat-list-item [ngClass]="'action-bar-title'" @growShrink matRipple #itemRipple="matRipple" (click)="itemRipple.disabled ? '' : toggleItemExpansion(action.id)">
```
mrmeku: Shouldn't this just be a class?

libs/ui/src/lib/build-status/build-status.component.scss
```diff
@@ -0,0 +1,49 @@
...
+    margin: 0 4px;
+  }
+
+  h3 mat-icon {
+    margin: 0 4px;
+  }
+}
+
+.inner-container {
+  height: calc(100% - 5px);
```
mrmeku: Why 5px? Mind extracting the value to a variable with a descriptive name?

libs/ui/src/lib/detailed-status/detailed-status.component.html
```diff
@@ -0,0 +1,35 @@
...
+      </mat-list>
+    </ng-container>
+  </div>
+  <mat-divider></mat-divider>
+  <div [ngClass]="'terminal-container'" fxFlex>
+    <ui-terminal [command]="command || commandResponse?.command" [outChunk]="commandResponse?.out"></ui-terminal>
+  </div>
+  <div class="toggle-container">
+    <button class="mat-icon-button" matTooltip="Toggle terminal output">
+      <mat-icon [ngClass]="rawOutputVisible ? 'output-icon-active' : ''" [svgIcon]="'console'" (click)="onRawOutputToggle()"></mat-icon>
```
mrmeku: svgIcon="console"

libs/ui/src/lib/detailed-status/detailed-status.component.scss
```diff
@@ -0,0 +1,45 @@
+@import '~@angular/material/theming';
+@include mat-core();
+
+$actions-palette: mat-palette($mat-grey, 100);
+$active-button-palette: mat-palette($mat-blue);
```
mrmeku: We already have a primary/error palette defined at the application level. Could you import those rather than redefining a slight variation of our primary palette?

libs/ui/src/lib/detailed-status/detailed-status.component.ts
```diff
@@ -0,0 +1,73 @@
...
+const COMMAND_REGEXP = /^\w+\s+([a-zA-Z0-9-_]+?)(\s+.*)?$/;
+
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: 'ui-detailed-status',
+  templateUrl: './detailed-status.component.html',
+  styleUrls: ['./detailed-status.component.scss'],
+  animations: [
+    trigger('growShrink', [
+      state('void', style({ width: '100vw' })),
```
mrmeku: When we switch to a more desktop friendly layout its likely that using 100vw/30vw won't work the way we want it to. Could you use '*' or a percentage instead?

libs/ui/src/lib/task-runner/task-runner.component.scss
```diff
@@ -29,22 +26,23 @@
   background-color: rgb(240, 240, 240);
 }
 
+.content-container {
+  height: calc(100% - 45px);
```
mrmeku: Can you extract 45px to a descriptive variable called $WINDOW_CHROME_HEIGHT and use it here and below?

libs/ui/src/lib/test-status/test-status.component.html
```diff
@@ -0,0 +1,68 @@
...
+        </mat-list-item>
+      </div>
+      <ng-container *ngIf="status?.errors.length > 0">
+        <mat-divider></mat-divider>
+        <h3 matSubheader>
+          Failed tests
+        </h3>
+        <mat-divider></mat-divider>
+        <mat-accordion>
+          <mat-expansion-panel *ngFor="let error of status?.errors">
```
mrmeku: You might consider adding a track by here so that if some subset of errors remains the the corresponding dom elements stick around

libs/ui/src/lib/test-status/test-status.component.html
```diff
@@ -0,0 +1,68 @@
...
+            <mat-expansion-panel-header>
+              <mat-panel-title [ngClass]="'error-label'">
```
mrmeku: just regular class

libs/ui/src/lib/build-status/build-status.component.html
```diff
@@ -0,0 +1,78 @@
...
+            </div>
+          </div>
+        </mat-list-item>
+      </div>
+      <ng-container *ngIf="status?.errors.length > 0">
+        <mat-divider></mat-divider>
+        <h3 matSubheader>
+          Errors
+        </h3>
+        <mat-list-item [ngClass]="'build-error'" *ngFor="let error of status?.errors">
```
mrmeku: Just a regular class

libs/ui/src/lib/detailed-status/detailed-status.component.html
```diff
@@ -0,0 +1,35 @@
+<div [ngClass]="'container'" fxLayout="row">
```
mrmeku: just a regular class

libs/ui/src/lib/detailed-status/detailed-status.component.html
```diff
@@ -0,0 +1,35 @@
...
+  <div [ngClass]="'details-container'" [ngSwitch]="getType()" [@growShrink]="rawOutputVisible ? 'shrink' : 'grow'">
```
mrmeku: just a regular class

libs/ui/src/lib/detailed-status/detailed-status.component.html
```diff
@@ -0,0 +1,35 @@
...
+          Task failed
+        </mat-list-item>
+        <mat-list-item *ngSwitchDefault>
+          Not started
+        </mat-list-item>
+      </mat-list>
+    </ng-container>
+  </div>
+  <mat-divider></mat-divider>
+  <div [ngClass]="'terminal-container'" fxFlex>
```
mrmeku: Just a regular class

package-scripts.js
```diff
@@ -46,7 +46,7 @@ module.exports = {
         write: 'prettier --write \"./server/**/*.ts\"',
         check: 'prettier --list-different \"./server/**/*.ts\"'
       },
-      test: 'node ./tools/scripts/test ./dist/server/test',
+      test: 'node ./tools/scripts/test ./server/test',
```
mrmeku: Oh sweet, does this mean we won't have to recompile to update server tests between runs?

server/src/api/detailed-status-calculator.ts
```diff
@@ -143,25 +235,125 @@ function parseBuildState(value: string) {
...
+    date,
+    time,
+    errors,
+    progress,
+    serverHost: serverMatch ? serverMatch[1] : state.serverHost,
+    serverPort: serverMatch ? Number(serverMatch[2]) : state.serverPort
+  };
+}
+
+const COLOR_OUTPUT_REGEXP = /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g;
```
mrmeku: So nasty, I love it

server/src/api/detailed-status-calculator.ts
```diff
@@ -143,25 +235,125 @@ function parseBuildState(value: string) {
...
+}
+
+function isFinishedProcessingChunks(value: string) {
+  return (
+    value.indexOf('Connected on socket') > -1 ||
+    (value.indexOf('chunk') > -1 && value.indexOf('Hash:') > -1)
+  );
+}
+
+type TestState = {
```
mrmeku: interface?

server/src/api/detailed-status-calculator.ts
```diff
@@ -143,25 +235,125 @@ function parseBuildState(value: string) {
...
+      failure,
+      success
+    };
+  } else {
+    return s;
+  }
+}
+
+function getNextTestErrorState(_buffer: TestError[], value: string) {
+  const newCaseRegExp = /^\s*[^\(\)]+\([^\)]+\)+(.+) FAILED\s*$/;
```
mrmeku: I LOVE IT

package-scripts.js
jaysoo: Yep that's correct!

libs/ui/src/lib/build-status/build-status.component.scss
jaysoo: Not needed, good catch! Probably something left in there when I was playing around with styling.

libs/feature-action-bar/src/lib/action-bar.component.html
jaysoo: Yup, good point!

jaysoo: Yup, good point!

.circleci/config.yml
jaysoo: Reverting config changes. Had some problems with yarn cache on circle, but let's see if it is fixed!

server/src/api/detailed-status-calculator.ts
jaysoo: 👍

.circleci/config.yml
jaysoo: Yup, config should be reverted. Had some caching issues on circle.

libs/ui/src/lib/task-runner/task-runner.component.scss
jaysoo: Good point!

libs/ui/src/lib/test-status/test-status.component.html
jaysoo: 👍

libs/utils/src/lib/command-runner.service.ts
```diff
@@ -95,6 +101,21 @@ export class CommandRunner {
               return of(c);
             }
           }),
+          distinctUntilChanged((a, b) => {
```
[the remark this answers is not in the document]
jaysoo: I found that new value was being emitted by the observable due to polling (even though nothing has changed), which is causing the UI to thrash.

server/src/api/command-normalizer.ts
```diff
@@ -0,0 +1,18 @@
+import { join } from 'path';
+
+export function normalize(cwd: string, cmds: string[]): string[] {
```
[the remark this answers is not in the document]
jaysoo: Some commands like `ng test ...` uses karma, which may have configurations that won't generate the expected output. This ensures that we can get what we need in the interim until build events.

libs/ui/src/lib/command-output/command-output.component.ts
```diff
@@ -0,0 +1,108 @@
...
+  @Input() command: string;
+  @Input() commandResponse: CommandResponse;
+  @Input() emptyTemplate?: TemplateRef<void>;
+  @Input() rawOutputVisibilityChange?: (isVisible: boolean) => void;
+
+  rawOutputVisible = false;
+  private readOutputLineCount = 0;
+
+  getType() {
+    const cmd =
```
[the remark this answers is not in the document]
jaysoo: 👍

libs/utils/src/lib/command-runner.service.ts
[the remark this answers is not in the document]
vsavkin: Since the UI should consume chunks, you can get the same looking chunk multiple times in a row, each of which will need to be applied to the terminal.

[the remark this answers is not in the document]
vsavkin: If we didn't use chunks, then it would have worked.

server/src/api/command-normalizer.ts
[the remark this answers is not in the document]
vsavkin: One of the design goals we follow is:
Running a command using Console should work the same way as running a command without console. Hence we cannot override karma configuration. The tests may not work or different tests may run.

The issue is actually broader:
There is no guarantee that  `ng test` will actually use karma. It can also use `jest` or whatever the user provisions in angular.json.

Given that, do you think it makes sense to get the builder name from angular.json to make sure `ng test` even uses karma before we start parsing the output?

PR #264, opened 2018-10-17
docs: add script to run electron app on Linux

dungahk:
It was missing from the CONTRIBUTING and people could think it was not possible


dungahk:
I followed the CONTRIBUTING guideline for the commit message, it says:

#### Type

The type must be one of the following:

* build
* feat
* fix
* refactor
* style
* docs
* test

Is that still valid or should that be updated as well?

dungahk:
Done!

PR #278, opened 2018-10-19
feat(workspaces): Use native file picker for workspace creation/imports

mrmeku:
Included in this change is a completely reworked UI for the workspaces component


libs/feature-workspaces/src/lib/new-workspace/new-workspace.component.scss
```diff
@@ -1,77 +1,32 @@
...
-    overflow: auto !important;
+  .path-form-field,
+  .path-form-field input {
+    cursor: pointer;
   }
 
-  .expand .mat-expansion-panel-content {
-    max-height: calc(100vh - 230px) !important;
+  .selected-schematic {
+    background: #dcdcdc !important;
```
jaysoo: It'd be good for us to define a set of app css variables, so we can keep things consistent. No need to do it now since it could be a cleanup task later.

libs/feature-workspaces/src/lib/new-workspace/new-workspace.component.ts
```diff
@@ -51,139 +48,106 @@ interface SchematicCollectionForNgNew {
...
-        null,
-        Validators.required,
-        makeNameAvailableValidator(finderService)
-      ),
-      collection: new FormControl(null, Validators.required)
-    });
+  handleSelection(event: MatSelectionListChange) {
+    if (event.option.selected) {
+      event.source.deselectAll();
+      event.option._setSelected(true);
```
jaysoo: Can we just use `event.option.selected = true`? Seems like a semi-private method that we may not want to rely on. Are we trying to avoid change from being reported?

libs/feature-workspaces/src/lib/new-workspace/new-workspace.component.ts
```diff
@@ -51,139 +48,106 @@ interface SchematicCollectionForNgNew {
...
+    if (this.ngNewForm.valid) {
+      const ngNewInvocation: NgNew.Variables = this.ngNewForm.value;
+      this.command = `ng new ${name} --collection=${
+        ngNewInvocation.collection
+      }`;
+      this.commandOutput$ = this.commandRunner
+        .runCommand(this.ngNewGQL.mutate(ngNewInvocation), false)
+        .pipe(
+          tap(command => {
+            if (command.status === 'successful') {
```
jaysoo: Could use `CommandStatus.SUCCESSFUL` instead from `@angular-console/utils`.

libs/feature-workspaces/src/lib/workspaces/workspaces.component.scss
```diff
@@ -11,6 +11,102 @@ mat-list-item {
   overflow: hidden;
 }
 
+mat-nav-list {
+  margin-bottom: 24px;
+}
+
+@media (max-width: 920px) {
```
jaysoo: Would love to pull these media queries into variables like `$small-screen`, `$medium-screen`, etc. Again, can do as a general cleanup chore later.

libs/feature-workspaces/src/lib/workspaces/workspaces.component.ts
```diff
@@ -22,13 +28,38 @@ import { WorkspacesService } from '../workspaces.service';
...
   }
+
+  ngOnInit() {
+    this.contextualActionBarService.breadcrumbs$.next([
+      { title: 'Select an Angular Workspace' }
+    ]);
+  }
+
+  importExistingWorkspace() {
+    this.workspacesService.selectExistingWorkspace().subscribe(result => {
```
jaysoo: Do we need a `pipe(first())` to make sure the stream completes... or manual unsubscribe on destroy?

libs/feature-workspaces/src/lib/new-workspace/new-workspace.component.ts
```diff
@@ -51,139 +48,106 @@ interface SchematicCollectionForNgNew {
...
-              name: 'Create',
-              disabled: new BehaviorSubject(formStatus !== 'VALID'),
-              invoke: this.createNewWorkspace$
-            }
-          ]
-        });
+  selectParentDirectory() {
+    this.workspacesService
+      .selectDirectoryForNewWorkspace()
+      .subscribe(result => {
```
jaysoo: Do we need an unsubscribe for this?

libs/feature-workspaces/src/lib/new-workspace/new-workspace.component.ts
mrmeku: Good call

libs/feature-workspaces/src/lib/workspaces/workspaces.component.ts
mrmeku: Added pipe to selectExistingWorkspace() itself

libs/feature-workspaces/src/lib/new-workspace/new-workspace.component.ts
mrmeku: Added pipe to selectDirectoryForNewWorkspace() itself

libs/feature-workspaces/src/lib/new-workspace/new-workspace.component.scss
mrmeku: At the very least I'm extracting the colors out to variables. We can move those variables to a common file in a clean up :)

libs/feature-workspaces/src/lib/new-workspace/new-workspace.component.ts
mrmeku: This hacky code is a workaround for https://github.com/angular/material2/issues/7157.

I've left a comment to reflect that

libs/feature-workspaces/src/lib/workspaces/workspaces.component.scss
mrmeku: Done!

PR #292, opened 2018-10-23
WIP: refactor: Use rounded material components

### LINKED ISSUE #290: UI-Flags component: Use rounded material components
When I was first creating the ui-flags component (the component which renders fields within a particular CLI schematic), Angular material did not yet have its "outline" style form fields (see https://stackblitz.com/edit/angular-uhbbyu?file=app/form-field-label-example.html for an example). Now that it does, we should use it instead of the custom rounded form fields that I hacked together initially. The benefits are not having brittle CSS that can break with angular material releases, using the built-in mat-hint component rather than absolutely positioning our hints per flag type and overall a better UI experience.

This will be an easy change for autocomplete-able, text-based and select flag types. A bit of custom work will need to be done for boolean types (maybe we should just represent these as dropdowns with two items or maybe we do something custom so they remain clickable toggles). 

In addition to using outline based form fields the UI forms could use some general UI touchups. Adding more whitespace to increase readability and potentially adding another column to the form for screens which are particularly wide. These and other tweaks can be left to whoever takes on this issue.

manekinekko:
Closes #290

### DIFF
--- a/libs/ui/src/lib/flags/flags.component.html
+++ b/libs/ui/src/lib/flags/flags.component.html
@@ -27,23 +27,25 @@
         fxLayout="row" fxLayoutGap="16px" [class.boolean]="field.type === 'boolean'" [class.text]="field.type !== 'boolean'">
         
         <ng-container *ngIf="field.type === 'boolean' && field.defaultValue !== null">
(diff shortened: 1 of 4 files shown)


manekinekko:
@mrmeku what do you about this change?


ezgif-5-38652e42816d

mrmeku:
Being continued in https://github.com/nrwl/angular-console/pull/310

PR #306, opened 2018-10-30
feat: adds settings page and fixes tests output

jaysoo:
- Test output will capture build errors now.
- New setting for enabling/disabling detailed status (defaults to true).


libs/feature-settings/src/lib/feature-settings.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
```
mrmeku: Remove this empty spec file or add some tests

libs/feature-settings/src/lib/feature-settings.component.ts
```diff
@@ -0,0 +1,16 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'angular-console-settings-feature-settings',
+  template: `
+    <p>
+      feature-settings works!
```
mrmeku: This looks not very functional?

server/src/schema/resolvers.ts
```diff
@@ -204,14 +204,18 @@ const Database: DatabaseResolvers.Resolvers = {
   },
   commands(_root: any, args: any) {
     try {
+      const settings = readSettings();
+      const includeDetailedStatus = settings.enableDetailedStatus;
       if (args.id) {
-        const c = commands.recent.find(c => c.id === args.id);
+        const c = commands.recent.find((c: any) => c.id === args.id);
```
mrmeku: Do we have an interface we can use in place of any?

libs/feature-settings/src/lib/feature-settings.component.ts
jaysoo: Will remove!

libs/feature-settings/src/lib/feature-settings.component.spec.ts
jaysoo: Good point

PR #310, opened 2018-11-02
feat(ui-flags): Use angular material rounded inputs for flags.

mrmeku:
Co-authored-by: Wassim Chegham <github@wassim.app>

### DIFF
--- a/libs/ui/src/lib/flags/flags.component.html
+++ b/libs/ui/src/lib/flags/flags.component.html
@@ -3,7 +3,8 @@
     <div class="headline">
       Configuration
     </div>
-    <mat-radio-group class="configuration-container" formControlName="configurations" name="configurations" fxLayout="row" fxLayoutGap="16px">
+    <mat-radio-group class="configuration-container" formControlName="configurations" name="configurations" fxLayout="row"
+      fxLayoutGap="16px">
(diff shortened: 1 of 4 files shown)


libs/ui/src/lib/flags/flags.component.html
```diff
@@ -12,9 +13,10 @@
   <mat-divider></mat-divider>
 
   <mat-accordion [multi]="true" displayMode="flat">
-    <mat-expansion-panel *ngFor="let fieldGroup of fieldGroups" [expanded]="fieldGroup.expanded" (opened)="resizeFlags.emit()" (closed)="resizeFlags.emit()">
+    <mat-expansion-panel *ngFor="let fieldGroup of fieldGroups" [expanded]="fieldGroup.expanded" (opened)="resizeFlags.emit()"
+      (closed)="resizeFlags.emit()">
       <mat-expansion-panel-header>
-        <mat-panel-title *ngIf="fieldGroups.length > 1" [ngSwitch]="fieldGroup.type" [ngClass]="'js-group-' + fieldGroup.type">
```
jaysoo: This class was used to drive e2e tests. They are commented out for now since they seem to be causing memory issues in CI, but something to keep in mind.

PR #318, opened 2018-11-06
feat: provide telemetry to console plugins

vsavkin:


libs/ui/src/lib/data-collection/data-collection.component.ts
```diff
@@ -8,7 +8,7 @@ import { AnalyticsCollector, Settings } from '@angular-console/utils';
 })
 export class DataCollectionComponent {
   constructor(
-    private readonly analytics: AnalyticsCollector,
+    @Inject('telemetry') private readonly telemetry: Telemetry,
```
mrmeku: Telemetry is a class. Why do you need the injection token?

vsavkin: we do. This is the only way for us to provide it to plugins. In the future, we can create some `angular-console-api` packages plugins can depend on, but we don't have anything like that right now.

Issue #320, opened 2018-11-06
Compilation errors break test detailed status


vsavkin:
Repro:

- Create a new CLI project
- Start running tests with --watch
- Introduce the compilation error by say, removing a quote (`describe('AppComponent, () => {`)
- You will see the error being displayed
- Fix the error
- The terminal will display the correct status (3 of 3 are successful), but the detailed status won't

jaysoo:
Looks like there was some unexpected output that clobbered the state.

Just tested locally with my fixes, and it seems to be working now.

Steps:

1. Run test with watch.
2. Tests all passed.
3. Introduce compile error.
4. See build failed.
5. Fix error.
6. Tests pass again.
7. Introduce expectation error.
8. See tests failed.
9. Fix tests.
10. See tests passed.

![test-run](https://user-images.githubusercontent.com/53559/48143602-25564400-e27d-11e8-88fd-6274d42ff788.gif)

PR #321, opened 2018-11-07
fix: Fixed build and test output for detailed status.

### LINKED ISSUE #319: Compilation errors aren't formatted correctly

jaysoo:
Closes #319, #320

- Formats build errors using <pre> so stack trace looks correct.
- Fixed test status when compliation error occurs.


libs/ui/src/lib/build-status/build-status.component.html
```diff
@@ -66,8 +66,8 @@
         <h3 matSubheader>
           Errors
         </h3>
-        <mat-list-item class="build-error" *ngFor="let error of status?.errors; trackBy: trackByError">
-          {{ error }}
+        <mat-list-item>
+          <pre>{{ status?.errors.join('\n') }}</pre>
```
mrmeku: Make sure to use `white-space: pre-wrap;` on this and other pre-tags you are rendering. Also make sure to set the font to `roboto-mono` (you could even do that in our global scss file

PR #325, opened 2018-11-08
feat(build): Update electron / render linux icons

mrmeku:


mrmeku:
Closes https://github.com/nrwl/angular-console/issues/212

PR #328, opened 2018-11-09
feat(ng-new): Allow users to set ng new schematic flags.

mrmeku:


mrmeku:
Closes https://github.com/nrwl/angular-console/issues/163 and https://github.com/nrwl/angular-console/issues/326

server/src/api/read-ngnews.ts
```diff
@@ -1,12 +1,27 @@
...
+        '@schematics-angular-ng-new-schema.json',
+        join(__dirname, '../../assets/')
+      )
     },
     {
       name: '@nrwl/schematics',
-      description: 'Enterprise-ready Angular CLI workspace'
+      description: 'Enterprise-ready Angular CLI workspace',
+      schema: readSchematicSchema(
+        '@nrwl-schematics-ng-new-schema.json',
```
vsavkin: I don't see `@nrwl-schematics-ng-new-schema.json`. Is it supposed to come with this PR as well?

server/src/schema/schema.graphql
```diff
@@ -122,7 +122,12 @@ type OpenDocResult {
 
 type Mutation {
   ngAdd(path: String!, name: String!): CommandStarted
-  ngNew(path: String!, name: String!, collection: String!): CommandStarted
+  ngNew(
+    path: String!
+    name: String
+    collection: String!
+    newCommand: [String!]!
```
vsavkin: doesn't have to be done in this PR, but it would be great to ! to other option lists in the schema.

server/src/api/read-ngnews.ts
mrmeku: Removed

server/src/schema/schema.graphql
mrmeku: Totally!

PR #413, opened 2018-11-25
New GraphQL Code Generator

kamilkisiela:
### New GraphQL Code Generator

- We changed the API a bit. It’s no longer required to use --flags with CLI, the codegen handles config files in js,yml and json.
- Better error messages
- We added a list of tasks that is visible in the terminal, to track the progress.
There’s no huge templates, everything is broken into smaller pieces, called plugins.
- And we have a website: https://graphql-code-generator.com
- and way more (we're going to release v1.0 and a blog post soon)

### Server
- It’s now possible to create and use an interface for the GraphQL Context in every piece of generated code, so you provide it once in the config file and the codegen takes care of rest keeping everything more strict than before.
- [A feature called Mappers](https://graphql-code-generator.com/docs/plugins/typescript-resolvers#mappers-overwrite-parents-and-resolved-values) that lets you map an interface to a GraphQL Type so whenever you want to resolve one it will be strongly typed and exactly what you expect it to be.

### Client
- now it generates multiple files in a single process and shares the GraphQL Schema
- it handles max 4 files at once, in parallel, so it should be way faster.

---

- [x] release and bump the `graphql-code-generator` to `v0.14.3`


mrmeku:
Could you re-base and re-gen the graphql.ts files? 

And I was thinking, whats your opinion on adding a the generated file to .gitignore? 
- The benefit would be to avoid this exact sort of merge collision
- The draw back would be a new contributor to the repo might wonder why their are imports to files which don't exist

What are your thoughts?

PR #414, opened 2018-11-26
WIP: feat: add support for desktop notifications

manekinekko:
fixes https://github.com/nrwl/angular-console/issues/164


manekinekko:
@vsavkin @mrmeku can you help explain how the settings are handled by the server?

libs/utils/src/lib/notification.service.ts
```diff
@@ -0,0 +1,27 @@
...
+export class Notification {
+  constructor(private readonly settings: Settings) {}
+
+  notify(body: string, title: string = 'Angular Console'): Observable<any> {
+
+    if (!this.settings.enableNotifications) {
+        return of({});
+    }
+
+    const notificationInstance = (window as any).Notification.requestPermission().then(
```
mrmeku: I think `@types/chrome` will have the typings for this api

mrmeku:
@manekinekko Here's how the server deals with settings: https://github.com/nrwl/angular-console/blob/master/server/src/api/read-settings.ts#L3 

That class has both a read and write method attached to it. 

Here is the api endpoint which the front end calls to update a user's preferences: 
https://github.com/nrwl/angular-console/blob/a9f4926b808fbf592286b875fea32a6f42245591/server/src/schema/resolvers.ts#L441

mrmeku:
I'm going to close this PR because it pretty stale now. Please re-open it when you get a chance to rebase!

Issue #418, opened 2018-11-29
fix: Recent GraphQL gen changes break server build on Windows


adammedford:
Building the source from master on Windows is failing as of right now.

Changing line 50 of package-scripts.js 
from: 
`'move-graphql': 'cp server/src/schema/schema.graphql dist/server/src/schema/schema.graphql',`
to: 
`'move-graphql': 'copy server\\src\\schema\\schema.graphql dist\\server\\src\\schema\\schema.graphql',`
allows Windows to build correctly. 

I'll submit a PR later if I have a chance. Just wanted to post here so others encountering this will have a workaround until there's an official solution.

PR #419, opened 2018-11-29
fix: make graphql gen script platform specific

adammedford:
Use 'withPlatform' in server.compile to gen Graphql with platform specific commands (copy vs cp)
#418


mrmeku:
Thanks for doing this!

Issue #304, opened 2018-10-29
[enhancement] Improve UX for file/module selection


adammedford:
First off - extraneous pandering. Awesome work guys - another awesome project. I was actually writing something along these lines myself when I discovered this project. Naturally, I had a few ideas for my implementations that  you've not added to your own yet.

The problem:
- Selecting modules as projects grow is tedious and error prone
- The CLI provides tab path completion - the Angular Console is currently restricted to dumb text boxes
- Angular CLI schematic schemas are made for the CLI and shouldn't be modified for GUI problems

The ask:
1.  Assume 'string' fields with names like 'directory', 'module', 'parentModule', should use different inputs (autocomplete + file modal) -  **I don't like this but I'm including for completeness**

2. Add a button next to 'string' fields that opens a dialog window to allow users to select from a list of options with recently used options being pushed to the top. It might be enough just to provide provide a file dialog typed to *.module.ts

I'm sure someone can come up with a better implementation but I'd be happy to offer a PR nonetheless.

adammedford:
So I was working on another project and it looks like there is an autocomplete already - it just isn't always used. For me it works on the vanilla Angular Schematics but I don't see any on the Nrwl schematics.

mrmeku:
@adammedford Could you make a small list of places you would expect to see the autocomplete? If you do that then I'll go through each one and add them :D.

adammedford:
@mrmeku I actually went ahead and made the changes.  I figured if I was going to make a list for you I might as well do it in the code on the off chance my formatting is acceptable. I'll submit a PR later tonight when I'm not blocked by a corporate proxy.

adammedford:
I'm hesitant to close this as this is really an ongoing thing. 

I think if we're going to continue hardcoding details about schematic extensions it'd be a good idea to add a section under the CONTRIBUTING.md providing a checklist of adding a new schematic library and outlining the logic Angular Console uses to interact with them. 

Perhaps if Angular Console continues to grow the schematics extension authors could be coaxed into providing an angular-console.json with the collection.json to provide additional hinting. Would be a lot more durable and up to date than the current approach.

Cammisuli:
We’ll probably come around and look at better autocomplete for additional fields. But I think what we have now is pretty good. We’re not going to force people to put specific json files to work with nx console. We’ll most likely use the nx nrwl/devkit to do more things. 

I’m going to close this because the initial ask was completed.

PR #420, opened 2018-11-29
fix: add new schematic collections for autocomplete and important fields

adammedford:
Add 'required' check for determining important schematic fields
Add 'parentModule' for '@nrwl/schematics' auto completions
Add '@angular/cdk', '@angular/material' to standard completion handling
Fixes a typo in methodname reoderFields => reorderFields

#304

### DIFF
--- a/libs/utils/src/lib/serializer.service.spec.ts
+++ b/libs/utils/src/lib/serializer.service.spec.ts
@@ -30,6 +30,7 @@ describe('Serializer', () => {
   });
 
   describe('normalizeSchematic', () => {
+    serializer = new Serializer();
     const schematic: Schematic = {
       collection: '@nrwl/testing',
       name: 'normalized',
@@ -67,9 +68,8 @@ describe('Serializer', () => {
         }
       ]
     };
-    let normalized: Schematic;
+    const normalized: Schematic = serializer.normalizeSchematic(schematic);
     it('should normalize the schematic, removing periods from the end of description', () => {
-      normalized = serializer.normalizeSchematic(schematic);
       expect(normalized.description.lastIndexOf('.')).toBeLessThan(
         normalized.description.length
       );
@@ -84,6 +84,12 @@ describe('Serializer', () => {
           case 'ignoreImportant':
             expect(field.important).toEqual(false);
             break;
+          default:
+            throw new Error(
+              `Unexpected field name '${
+                field.name
+              }' in normalizeSchematic test cases`
+            );
         }
       });
     });
--- a/libs/utils/src/lib/serializer.service.spec.ts
+++ b/libs/utils/src/lib/serializer.service.spec.ts
@@ -1,3 +1,4 @@
+import { Schematic, Field } from '@angular-console/schema';
 import { Serializer } from './serializer.service';
 
 describe('Serializer', () => {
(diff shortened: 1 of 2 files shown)


libs/utils/src/lib/serializer.service.ts
```diff
@@ -10,12 +10,13 @@ export class Serializer {
       ...f,
       important:
         f.positional ||
+        f.required ||
```
adammedford: Some fields are 'required' by Angular Console while being buried in the 'Optional Fields' panel. 

I added this because I think this was an oversight - but if it was intentional than maybe 'Optional' should be renamed in the UI so as not to be misleading.

libs/utils/src/lib/serializer.service.ts
```diff
@@ -114,15 +115,19 @@ export class Serializer {
...
-    if (collection === '@nrwl/schematics' && name === 'module') {
-      return 'absoluteModules';
-    }
-    if (
-      collection === '@schematics/angular' ||
-      collection === '@ngrx/schematics'
-    ) {
-      if (name === 'project') return 'projects';
-      if (name === 'module') return 'localModules';
+    switch (collection) {
```
adammedford: As the list of collections is growing I thought a switch was a bit more readable here.

libs/utils/src/lib/serializer.service.ts
mrmeku: Nope, you nailed it. That was a bad over sight. If you wouldn't mind could you add a regression test to make sure it doesn't get into a bad state again?

libs/utils/src/lib/serializer.service.ts
mrmeku: Definitely!

mrmeku:
@adammedford  This change looks great! If you wouldn't mind could you add a regression test to the logic for deeming a field important? That would be amazing if you could.

adammedford:
> @adammedford This change looks great! If you wouldn't mind could you add a regression test to the logic for deeming a field important? That would be amazing if you could.

Absolutely, will follow-up in a bit. Thanks for taking a look at this.

libs/utils/src/lib/serializer.service.spec.ts
```diff
@@ -27,4 +28,64 @@ describe('Serializer', () => {
...
+    };
+    let normalized: Schematic;
+    it('should normalize the schematic, removing periods from the end of description', () => {
+      normalized = serializer.normalizeSchematic(schematic);
+      expect(normalized.description.lastIndexOf('.')).toBeLessThan(
+        normalized.description.length
+      );
+    });
+    it('should recognize positional and required fields as important but ignore previous values of important', () => {
+      normalized.schema.forEach((field: Field) => {
```
mrmeku: You've written your tests in such a way that the order they are executed in matters. Could you instead use a beforeEach function which instantiates `normalized` rather than relying on the previous test?

libs/utils/src/lib/serializer.service.spec.ts
```diff
@@ -27,4 +28,64 @@ describe('Serializer', () => {
...
+        switch (field.name) {
+          case 'required':
+          case 'positional':
+            expect(field.important).toEqual(true);
+            break;
+          case 'ignoreImportant':
+            expect(field.important).toEqual(false);
+            break;
+        }
```
mrmeku: Add a default here which throws and makes the test fail

libs/utils/src/lib/serializer.service.spec.ts
adammedford: I'm going to lead with "I didn't do the specific thing you asked me to do." I accomplished the same goal, but I didn't want to use a beforeEach to instantiate the result of a method call. I didn't want to have it quietly increase test execution time as new test cases are added when we really only care about testing that one value.

Issue #426, opened 2018-11-30
[bug] NgRx extensions don't add schematics despite showing 'Installed'


adammedford:
Two extensions are listed in the extensions lists 
1. @ngrx/effects
2. @ngrx/store

Neither actually adds schematics because the @ngrx/schematics is the correct package - but not listed.

The other two have an 'ng-add' schematic that isn't used for whatever reason.

mrmeku:
Thanks for the details Adam. I'll fix this for the next release

PR #427, opened 2018-11-30
 fix: add missing '@ngrx/schematics' to extensions

adammedford:
@ngrx/schematics appears to be the actual package for NgRx
@ngrx/effects + @ngrx/store didn't actually add schematics to the list despite being installed


adammedford:
#426

adammedford:
Just realized that I duplicated my previous changes to package-scrips.js. Will fix later - I'm limited to the web interface right now so rebasing to resubmit isn't an option.

server/src/api/read-extensions.ts
```diff
@@ -31,6 +31,10 @@ export function availableExtensions() {
       name: '@nativescript/schematics',
       description: 'NativeScript Support'
     },
+    {
+      name: '@ngrx/schematics',
+      description: 'NgRx state management generators'
+    },
     { name: '@ngrx/effects', description: 'NgRx side effect management' },
```
adammedford: I left the other extensions in here since they each have a single schematic that I'm guessing is poorly formed as they do not display in the list.

adammedford: Will do it when I get home. I saw after I submitted this that canAdd removes schematics named 'ng-add'

PR #431, opened 2018-12-03
fix: add missing '@ngrx/schematics' to extensions

adammedford:
Add actual NgRx extension with schematics
Remove NgRx packages that don't


adammedford:
#426 #427 @mrmeku

PR #434, opened 2018-12-07
fix(schematics): Filter out hidden/private schematics

mrmeku:


server/src/api/read-schematic-collections.ts
```diff
@@ -90,9 +90,9 @@ function readSchematicCollections(
...
 }
 
-function canAdd(
+export function canAdd(
   name: string,
-  s: { hidden: boolean; extends: string; schema: string }
+  s: { hidden: boolean; private: boolean; schema: string }
 ): boolean {
-  return !s.hidden && !s.extends && !!s.schema && name !== 'ng-add';
+  return !s.hidden && !s.private && Boolean(s.schema) && name !== 'ng-add';
```
mrmeku: @vsavkin I removed the check of !s.extends because that seemed wrong. Could you comment and make sure that me removing it is not a bad thing?

mrmeku: Cool, I'll leave it in

PR #436, opened 2018-12-07
fix: Navigate back to previous location when active target or schematic is not present.

### LINKED ISSUE #435: Closing task runner should go back to previous screen
When the user is on the Project screen and jumps to an action on the right, it brings up the task runner.

After the user closes the task runner, we do not bring them back to where they came from, but instead they either land on `Generate Code` or `Run Tasks`.

screen shot 2018-12-07 at 11 28 19 am

jaysoo:
**Assumption:** There is always a location to go back to (which is true for this desktop app)

**Note:** There is a small delay in updating sidenav as seen below.

tasks

- Closes #435

### DIFF
--- a/apps/angular-console-e2e/src/integration/projects.spec.ts
+++ b/apps/angular-console-e2e/src/integration/projects.spec.ts
@@ -31,4 +31,15 @@ describe('Projects', () => {
       expect($p[0].value).to.equal('proj');
     });
   });
+
+  it('provides navigation to and from command runners', () => {
+    cy.contains('Generate Component').click();
+    cy.get('.exit-action').click();
+
+    projectNames($p => {
+      expect($p.length).to.equal(2);
+      expect(texts($p)[0]).to.contain('proj');
+      expect(texts($p)[1]).to.contain('proj-e2e');
+    });
+  });
 });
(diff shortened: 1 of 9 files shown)


jaysoo:
@mrmeku When you get a chance, I made some changes so that sidebar navigation isn't delayed when you close the task selector.

libs/utils/src/lib/location-ext.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  constructor(
+    private readonly router: Router,
+    private readonly location: Location
+  ) {}
+
+  ngOnDestroy() {
+    this.additionalNavigationSubscription.unsubscribe();
+  }
+
+  isInitialRoute() {
```
vsavkin: This and additionalNavigationSubscription are only used in tests, right?

Should they be removed?

libs/utils/src/lib/location-ext.service.ts
```diff
@@ -0,0 +1,44 @@
+import { Injectable, OnDestroy } from '@angular/core';
+import { Router, NavigationEnd, NavigationExtras } from '@angular/router';
+import { filter, skip, take } from 'rxjs/operators';
+import { Location } from '@angular/common';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class LocationExtService implements OnDestroy {
```
vsavkin: I think "RouterNavigationService" is better. The service doesn't extend the location service. The only public method has the signature of `router.navigate`.

PR #447, opened 2018-12-21
feat: make it easier to see what extensions are installed

georgewilde:
The extensions screen makes it clear to see which extensions are installed.


vsavkin:
@georgewilde thank you

Issue #156, opened 2018-08-15
Read schematic defaults from angular.json


Brocco:
The Angular CLI provides the way to default values for schematics, my request is to parse these values into the console as defaults.

An example for this would be the default CSS preprocessor extension to use when generating components.

mrmeku:
@vsavkin If we were to do this, in your opinion would we want to default to having the flags present in the generate command or rather, just update the default in the command metadata so that the help text / UI renders the correct default value without adding flags? I don't have a strong opinion on which is better so I'll leave it to you to decide (or anyone else who has a strong opinion of course)

PR #450, opened 2018-12-21
fix: Use defaults from angular.json in generate form

isaacplmann:
Fix for #156 .

This takes the approach of overriding the `defaultValue` when the schema is normalized.
screen shot 2018-12-21 at 15 26 58

Another possible solution would be to keep the existing `defaultValue` and add a `projectDefault`, so that the initial value is set to the `projectDefault`, but the `defaultValue` from the schema definition is still displayed in the description.
image

### DIFF
--- a/apps/angular-console-e2e/src/integration/generate.spec.ts
+++ b/apps/angular-console-e2e/src/integration/generate.spec.ts
@@ -58,7 +58,14 @@ describe('Generate', () => {
 
     const name = uniqName('example');
     cy.get('input[name="name"]').type(name);
-    cy.get('input[name="project"]').type('proj');
+    cy.get('input[name="project"]').type('proj{esc}');
+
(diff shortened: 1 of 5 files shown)


vsavkin:
it looks good. Could you add an e2e test or modify an existing one?

mrmeku:
@isaacplmann Any update on this PR?

isaacplmann:
I plan to finish it up on Monday.

isaacplmann:
Added an e2e test.

Notes:
- `read-schematic-collections.ts` has several functions passing around the workspaceDefaults, which isn't ideal, but the only other option is to pass down the root baseDir.
- In `setup-e2e-fixtures.js`, I manually set the `schematics` property in `tmp/proj/angular.json` so that we have a default workspace value to test against.

mrmeku:
Its passing now. Rebase and merge!

PR #452, opened 2018-12-27
chore: automatically format using prettier when commiting

jaysoo:
Automatically runs `yarn start format` before commits. This should prevent any failures during formatting checks in CI (unless you skip git hooks).

Also, the `frontend.format` was formatting server code (and tmp code from E2E tests) as well. I restricted the pattern to only `apps` and `libs`.

@mrmeku What do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "@nrwl/nx": "7.4.0"
   },
   "devDependencies": {
-    "shx": "^0.3.2",
     "@angular-devkit/build-angular": "0.11.4",
     "@angular-devkit/build-ng-packagr": "0.11.4",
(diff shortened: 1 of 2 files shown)


vsavkin:
Looks good. We'll merge it once the vscode PR gets merged.

mrmeku:
Needs a rebase, then lets merge it!

package.json
```diff
@@ -187,5 +188,10 @@
     "testMatch": [
       "**/*.+(spec|test).(ts|js)"
     ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn start format"
```
mrmeku: Do you need to do a git add . afterwards?

PR #459, opened 2019-01-07
fix: tweak polling constants to improve perf

vsavkin:


libs/server/src/lib/utils.ts
```diff
@@ -287,14 +287,14 @@ export function normalizePath(value: string): string {
...
+ * both the data sets get updated every 60 seconds.
  */
 export function cacheFiles(p: string) {
   setTimeout(() => {
     files[p] = listFiles(p);
     fileContents = cacheJsonFiles(p);
     setTimeout(() => {
       cacheFiles(p);
-    }, 30000);
+    }, 60000);
```
mrmeku: Can you make this a constant like our other polling constants?

PR #476, opened 2019-01-17
fix(vscode): Do not add commands to bash history

### LINKED ISSUE #468: All commands in system console history
All commands that have been processed remain in the console history (try `history` in system terminal)

image

mrmeku:
Fixes #468


mickz18:
Thanks, will fix also problem with non default posix terminal like Fish

PR #478, opened 2019-01-17
feat: pinned projects

isaacplmann:
Allow pinning on the projects page

1. When the component is loaded, the `pinnedProjectNames` are loaded from the settings service.  Any modifications are saved locally in the component and updated in settings.  Reloading the component will load the graphql settings version.  

    **Alternative:**  I could make the settings service return an Observable of the workspace settings to avoid having any local component state.

2. I tried copying the `@growShrink` animation from the `workspaces` component, but that does not appear to play nicely with `<cdk-virtual-scroll-viewport>`

### DIFF
--- a/libs/feature-workspaces/src/lib/projects/projects.component.ts
+++ b/libs/feature-workspaces/src/lib/projects/projects.component.ts
@@ -77,7 +77,6 @@ export class ProjectsComponent implements OnInit {
           return {
             ...p,
             actions: this.createActions(p)
-            // pinned: this.pinnedProjectNames.includes(p.name)
           };
         });
         return projects;
--- a/libs/feature-workspaces/src/lib/projects/projects.component.ts
+++ b/libs/feature-workspaces/src/lib/projects/projects.component.ts
@@ -2,11 +2,12 @@ import { ChangeDetectionStrategy, Component, OnInit } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { Project } from '@angular-console/schema';
 import { combineLatest, Observable, of } from 'rxjs';
-import { map, startWith, switchMap, shareReplay } from 'rxjs/operators';
+import { map, startWith, switchMap, shareReplay, tap } from 'rxjs/operators';
 import {
   PROJECTS_POLLING,
   Settings,
-  CommandRunner
+  CommandRunner,
+  toggleItemInArray
 } from '@angular-console/utils';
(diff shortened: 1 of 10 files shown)


libs/feature-workspaces/src/lib/projects/projects.component.ts
```diff
@@ -60,8 +75,17 @@ export class ProjectsComponent implements OnInit {
       }),
       map((r: any) => {
         const w = r.data.workspace;
-        const projects = w.projects.map((p: any) => {
-          return { ...p, actions: this.createActions(p) };
+        const workspaceSettings = this.settings.getWorkspace(
+          this.workspacePath
+        );
+        this.pinnedProjectNames =
```
mrmeku: Should you maybe use a set instead of list? Or are you using the ordering of the list for presentation?

mrmeku: Also, did you consider making pinnedProjectNames an observable? Just wondering (I usually find observables to be more reliable with Angular's change detection)

libs/server/src/schema/schema.graphql
```diff
@@ -220,6 +220,7 @@ type WorkspaceDefinition {
   path: String!
   name: String!
   favorite: Boolean
+  pinnedProjectNames: [String]
```
mrmeku: [String!]! is preferred

isaacplmann:
I believe this PR is ready to be merged

isaacplmann:
@mrmeku

libs/feature-workspaces/src/lib/projects/projects.component.ts
```diff
@@ -60,8 +68,17 @@ export class ProjectsComponent implements OnInit {
...
+          (workspaceSettings && workspaceSettings.pinnedProjectNames) || [];
+        const projects: Project[] = w.projects.map((p: Project) => {
+          return {
+            ...p,
+            actions: this.createActions(p)
+            // pinned: this.pinnedProjectNames.includes(p.name)
```
mrmeku: Can you remove this line if its not being used?

libs/server/src/assets/schema.graphql
```diff
@@ -219,6 +219,7 @@ type WorkspaceDefinition {
   path: String!
   name: String!
   favorite: Boolean
+  pinnedProjectNames: [String!]
```
mrmeku: Can you do [String!]! instead of just [String!] ?

isaacplmann:
@mrmeku Changes applied

PR #480, opened 2019-01-18
fix(readme): Styling core team table.

vetom:
This PR fixes the Core Team table also I change the images inline-style to the reference-style to make it easier to maintain.


mrmeku:
Thank you so much!

PR #481, opened 2019-01-18
feat(build): Bring vscode and electron builds more inline

mrmeku:


mrmeku:
image

Refactored package scripts enumeration

PR #490, opened 2019-01-29
feat: Show all actions from project view

isaacplmann:


libs/server/src/lib/api/read-settings.ts
```diff
@@ -1,10 +1,9 @@
 import { Store } from '@nrwl/angular-console-enterprise-electron';
 import { Settings } from '../generated/graphql-types';
 
-/* tslint:disable */
 export function readSettings(store: Store): Settings {
   const settings: Settings = store.get('settings') || {};
-  if (settings.canCollectData === undefined) {
+  if ((settings as any).canCollectData === undefined) {
```
mrmeku: This typecast worries me. Why is is necessary?

libs/server/src/lib/api/read-recent-actions.ts
```diff
@@ -0,0 +1,34 @@
...
+  const existingActions = readRecentActions(store, projectPath);
+  store.set(getRecentActionsKey(projectPath), [
+    { actionName, schematicName },
+    ...existingActions
+      .filter(
+        action =>
+          action.actionName !== actionName ||
+          action.schematicName !== schematicName
+      )
+      .slice(0, MAX_RECENT_ACTIONS - 1)
```
mrmeku: You don't need the - 1 here since the upper bound of slice is not included

libs/schema/src/index.ts
```diff
@@ -76,6 +76,12 @@ export interface Project {
   projectType: string;
   root: string;
   architect: Builder[];
+  recentActions: RecentAction[];
+}
+
+export interface RecentAction {
```
mrmeku: Our graphql type generators should make this interface for you. Use those generated interfaces so they are kept inline with our schema.

libs/feature-workspaces/src/lib/projects/projects.component.ts
```diff
@@ -37,31 +61,64 @@ export class ProjectsComponent implements OnInit {
...
+                })
+              )
+            )
+        );
       }),
-      map((r: any) => {
-        const w = r.data.workspace;
-        const projects = w.projects.map((p: any) => {
-          return { ...p, actions: this.createActions(p) };
+      filter(([r1, r2]) => !!r1 && !!r2),
```
mrmeku: Boolean(r1 && r2) is a bit clearer in my opinion

libs/server/src/lib/api/read-settings.ts
isaacplmann: The `Settings` type has `canCollectData` as required, which makes this line fail linting because it is always false.  My guess is the data model was changed at some point and this check needs to be there for old cached values.  I could disable tslint on that line instead of casting to `any` if you prefer.

libs/server/src/lib/api/read-recent-actions.ts
isaacplmann: I need the - 1 because I'm adding `{ actionName, schematicName }` at the front and want to remove the last element (assuming we've reached the `MAX_RECENT_ACTIONS`)

isaacplmann: I'll move the slice outside the `]` and take off the `- 1` so that I don't confuse future readers

libs/schema/src/index.ts
isaacplmann: Ok

libs/feature-workspaces/src/lib/projects/projects.component.ts
isaacplmann: Ok

libs/server/src/lib/api/read-settings.ts
isaacplmann: Switched to disabling tslint on that line

isaacplmann:
@mrmeku Ready for review

mrmeku:
@isaacplmann This needs to be rebased

isaacplmann:
@mrmeku not sure why the test is failing on CI.

Locally, I ran:

`yarn start prepare.e2e && yarn start e2e.headless`

And it succeeded.

PR #494, opened 2019-01-31
fix(frontend): close task should always go to the previous route

vsavkin:


libs/utils/src/lib/location-ext.service.ts
```diff
@@ -2,29 +2,36 @@ import { Injectable, OnDestroy } from '@angular/core';
...
-      this.isFirstRoute = false;
-    });
+  additionalNavigationSubscription: Subscription;
 
   constructor(
     private readonly router: Router,
     private readonly location: Location
-  ) {}
+  ) {
+    console.log('Created extended location');
```
mrmeku: Remove the console log?

PR #513, opened 2019-02-06
fix(new-workspace): Create a sandbox to create new workspaces

mrmeku:


package.json
```diff
@@ -77,12 +76,14 @@
     ]
   },
   "dependencies": {
-    "@angular/cli": "7.3.0",
-    "@schematics/angular": "7.3.0",
-    "@nrwl/schematics": "7.5.1",
+    "semver": "5.6.0",
```
vsavkin: I think all of them should be devDependencies.

PR #514, opened 2019-02-06
fix: number of columns gets reset which breaks the output

vsavkin:


libs/server/src/lib/resolvers/query.resolver.ts
```diff
@@ -176,3 +178,13 @@ function serializeCommand(
         : null
   };
 }
+
+function serializeCommandInList(c: CommandInformation): any {
+  return {
+    id: c.id,
+    type: c.id,
```
mrmeku: Why is the type the ID? Can you maybe leave a comment to explain that logic since its non-intuitive?

PR #557, opened 2019-02-21
feat(output): Adds visualization for serve and build output.

jaysoo:
## Changes

- Adds stats detail to build and serve tasks.
  - Errors and warnings are displayed as "Problems" (if they exist).
- Fixed type error in `npmscript.component.ts`.

## Build

screen shot 2019-02-21 at 1 22 10 pm

## Serve

Note that serve stats are very limited since we do not have a way to inspect the webpack stats object. We do have some modules and assets information based on terminal output.

screen shot 2019-02-21 at 1 19 17 pm

screen shot 2019-02-21 at 1 09 25 pm


libs/server/src/lib/api/detailed-status-calculator.ts
```diff
@@ -527,3 +549,13 @@ function collectPreviousErrorsFrom(idx: number, lines: string[]): string[] {
 
   return xs;
 }
+
+function parseTime(s: string) {
+  if (s.endsWith('ms')) {
```
mrmeku: What happens otherwise? Throw an error? Log an error? Something?

jaysoo: Just return the string that comes out put `ng build` or `ng serve`. I just want to format it to a nice string like  `1.02s`, but it's fine if it remains ugly as well. :)

PR #572, opened 2019-03-12
perf(projects): Improve rendering performance

mrmeku:


isaacplmann:
I don't think this is what you wanted:
image

Also, I don't see the animations working on the projects screen.

mrmeku:
@isaacplmann I removed the animations (they were broken). This is ready for another look now that its less buggy

isaacplmann:
@mrmeku I think I was doing something wrong when I was spinning up the PR to view in the browser.  It looks better now, but i can't get the recent actions list to ever change.

isaacplmann:
It looks like the recent actions are only broken for `test` and `lint` actions.  But it wasn't caused by this PR.

PR #573, opened 2019-03-12
feat(output): Adds webpack analyzer

jaysoo:


libs/server/src/lib/utils/stats.ts
```diff
@@ -147,7 +165,7 @@ export function calculateStatsFromChunks(cs: Chunk[]) {
 /* ------------------------------------------------------------------------------------------------------------------ */
 
 interface ChunkData {
-  id: number;
+  id: number | string;
```
mrmeku: Why both?

libs/ui/src/lib/build-status/build-status.component.html
```diff
@@ -147,6 +147,48 @@ <h3>Problems</h3>
...
+        </div>
+        <div class="actions" fxLayout="row" fxLayoutGap="20px">
+          <mat-select
+            matRipple
+            fxFlex
+            [value]="currentChunkId$ | async"
+            (selectionChange)="handleChunkFileSelection($event)"
+          >
+            <mat-option
+              *ngFor="let chunk of (chunks$ | async)"
```
mrmeku: Is it worth it to use a trackBy here?

libs/ui/src/lib/modules-graph/modules-graph.component.scss
```diff
@@ -0,0 +1,47 @@
...
+}
+
+.svg-container {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+.sequence {
+  width: 600px;
```
mrmeku: Can we make this more responsive? Why 600px?

libs/ui/src/lib/modules-graph/modules-graph.component.ts
```diff
@@ -0,0 +1,245 @@
+import { Component, ElementRef, Input, ViewChild } from '@angular/core';
+import { select, selectAll } from 'd3-selection';
+import { arc } from 'd3-shape';
+import { hierarchy, partition } from 'd3-hierarchy';
+import 'd3-transition';
+import { FormatFileSizePipe } from '../format-file-size.pipe';
+
+interface Chunk {
```
mrmeku: Consider running much of the logic within this component outside of the angular zone. D3 is known for taking a huge performance hit when running inside of a zone

libs/server/src/lib/utils/stats.ts
jaysoo: webpack comes back with both... but i can normalize to a string.

libs/ui/src/lib/modules-graph/modules-graph.component.scss
jaysoo: It becomes somewhat hard to use when too big or too small. Could consider a min-max range maybe?

libs/server/src/lib/utils/stats.ts
mrmeku: Do it!

apps/angular-console-e2e/src/integration/tasks.spec.ts
```diff
@@ -244,7 +244,7 @@ describe('Tasks', () => {
     cy.writeFile('../../tmp/proj/src/app/app.component.spec.ts', PASSING_TESTS);
   });
 
-  it('runs serve task', () => {
+  it.only('runs serve task', () => {
```
mrmeku: You don't really want this

mrmeku: You don't really want this

PR #577, opened 2019-03-13
fix: Handle Windows paths when building webpack viz.

jaysoo:
Also adds colours to apps and libs + styling tweaks.


libs/ui/src/lib/modules-graph/modules-graph.component.ts
```diff
@@ -14,14 +14,17 @@ import { FormatFileSizePipe } from '../format-file-size.pipe';
 
 interface Chunk {
   id: string;
-  name: string;
+  name: void | string;
```
mrmeku: void?

jaysoo: `undefined`, i mixed it up with Flowtype

PR #607, opened 2019-03-25
feat(IDE): Initial push for webstorm plugin

fkolar:
- Currently it can start / stop npm process
- It provides RPC channel to create communication link between java and JS

    For more info please check the README.md as DEV-INFO.md. We should keep it extra branch !


mrmeku:
We wont be able to commit the jxbrowser jar files as they increase the repo size by more than 10x
Can we use a dependency manager here?

PR #616, opened 2019-04-05
fix(intellij): exclude *.zip file from the buildPlugin process

fkolar:
This make sure it does not include `Angular Console-X.X.X.zip `from `dist/apps/intellij `into the gradle build process `buildPlugin`.


fkolar:
@mrmeku this is the thing we talked about to exclude plugin bundle that could be already under dist/apps/install if you previously run nps package.intellij

PR #626, opened 2019-04-06
fix(README.md): Fixing Frank's Github link under core contributors

fkolar:


fkolar:
@mrmeku one more small adjustment

PR #638, opened 2019-04-11
chore: uses in-memory PlatformLocation when in vscode

jaysoo:
For all environment except dev, we will use an in-memory version of `PlatformLocation` (e.g. doesn't sync with `window.location` object. This will allow use to use the console in platforms that do not support `window.location` -- such as vscode.

TODO: We still need to switch out of iframe for vscode, but in theory this works.


apps/angular-console/src/app/app.module.ts
```diff
@@ -77,6 +79,34 @@ export function initApollo(
...
+  },
+  { provide: ENVIRONMENT, useValue: environment as Environment },
+  { provide: IS_VSCODE, useValue: environment.application === 'vscode' },
+  { provide: IS_INTELLIJ, useValue: environment.application === 'intellij' },
+  { provide: IS_ELECTRON, useValue: environment.application === 'electron' }
+];
+
+declare const window: any;
+
+if (window.__USE_IN_MEMORY_LOCATION__) {
```
mrmeku: Just add this to the environment.providers field.

apps/vscode/src/app/webview.factory.ts
```diff
@@ -90,6 +90,10 @@ export function getIframeHtml(iframeUrl: string) {
           function onIframeLoad() {
             document.body.querySelector('iframe').classList.add('fade-in');
           }
+          
+          // This will be used in the client app to switch to memory location instead of the browser version.
+          // See: apps/angular-console/src/app/app.module.ts
+          window.__USE_IN_MEMORY_LOCATION__ = true;
```
mrmeku: remove this

libs/utils/src/lib/in-memory-platform-location.service.ts
```diff
@@ -0,0 +1,90 @@
+import {
+  DOCUMENT,
+  LocationChangeEvent,
+  LocationChangeListener,
+  PlatformLocation
+} from '@angular/common';
+import { Inject, Injectable, NgZone } from '@angular/core';
+import { ɵgetDOM as getDOM } from '@angular/platform-browser';
+import { Subject } from 'rxjs';
+import * as url from 'url';
```
mrmeku: Can you import { parse } alone (for the tree shaking)?

libs/utils/src/lib/in-memory-platform-location.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  pushState(state: any, title: string, newUrl: string) {
+    this.replaceState(state, title, newUrl);
+  }
+
+  forward() {
+    throw new Error('Not implemented');
+  }
+
+  back() {
+    throw new Error('Not implemented');
```
mrmeku: Reminds me of when I threw my back out

PR #670, opened 2019-04-25
chore: organize scripts by app instead of target kinds.

jaysoo:
Updates package scripts and contributing docs.

**Changes**
- Commands like `prepare.vscode` and  `package.vscode` are now `vscode.prepare` and  `vscode.package` respectively.
- The dev and e2e commands are mostly intact from usage POV: `dev.up`, `e2e.prepare`, `e2e.up`, etc.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
           paths:
             - ~/.cache
       - run: yarn start format.and.lint.check
-      - run: yarn start prepare.e2e
+      - run: yarn start e2e.prepare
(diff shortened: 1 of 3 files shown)


vsavkin:
@jaysoo 

1. since you are redoing the scripts, could you make sure the commands outlined in contributing work. For instance, currently we have `prepare.electron` and `prepare.dev.electron`. The second one isn't mentioned in `contributing.md`. 
2. could you also make sure that when developing in the browser, the location gets updated. I always have to patch env files to make it work.

PR #717, opened 2019-06-01
style: Change container height to show last task item

peterblazejewicz:
The last item is hidden below the fold and not accessible to
user.

Screenshot 2019-06-01 at 23 16 00

Thanks!


mrmeku:
Thanks so much for the PR!

PR #734, opened 2019-06-29
fix: Use build options from project workspace configuration

### LINKED ISSUE #605: [vscode]: Defaults in production build
After upgrading to the latest version of the extension (**7.4.3**) we have noticed that the production build command has `sourceMap` set to `true` and `statsJson` enabled by default which was not in the previous versions.

Is this intentional? If yes, is there a way to override them instead of changing them manually on each build? In our case, we have an automation that uploads the whole `dist` directory automatically and we don't want to deploy stats.json and source maps on the production server.

bampakoa:
The build command of a project applies default values from workspace configuration first. Fixes #605

### DIFF
--- a/libs/feature-run/src/lib/target/target.component.html
+++ b/libs/feature-run/src/lib/target/target.component.html
@@ -7,7 +7,7 @@
     [options]="project.architect[0].options"
     [prefix]="getPrefix(project.architect[0].name, project.name)"
     [fields]="project.architect[0].schema"
-    [init]="initSourceMapAnsStatsJson()"
+    [init]="initSourceMapAnsStatsJson(project.architect[0])"
     [runSyntax]="runSyntax(project.architect[0].name)"
     (value)="onFlagsChange($event)"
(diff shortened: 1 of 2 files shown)


bampakoa:
I tried to run :

- `yarn start test`
- `yarn start e2e.prepare`
- `yarn start e2e.up`

according to the contributions guidelines but I got some errors that does not seem to be related with the fixed code. The first one fails on `stats utils › generateStats › returns stats with sourcemap`, the second one does not running at all and the third one has error that is related with the title of the application

```
CypressError: Timed out retrying: Expected to find content: 'Extensions' within the selector: 'div.title' but never did.

Because this error occurred during a 'before each' hook we are skipping the remaining tests in the current suite: 'Extensions'
```

Please bare in mind that the Cypress error appears in all sections: Extensions, Forms, Projects etc...

Is there something that I could do for all these?

mrmeku:
Thanks so much for your PR! I'll get it merged on Monday

PR #750, opened 2019-07-16
fix: Reduce logging noise and handle schematics without a description

mrmeku:


mrmeku:
Fixes https://github.com/nrwl/angular-console/issues/722
Addresses parts of https://github.com/nrwl/angular-console/issues/740

PR #751, opened 2019-07-16
fix(vscode): Remove logging from angular-console-enterprise-electron

mrmeku:


mrmeku:
Fixes https://github.com/nrwl/angular-console/issues/740

PR #792, opened 2019-08-09
feat(vscode): Use vscode tasks

mrmeku:


mrmeku:
Fixes https://github.com/nrwl/angular-console/issues/681

PR #800, opened 2019-08-19
feat: default nest port changed

### LINKED ISSUE #569: Crash when serving default NestJS app
Issue: Angular Console crashes after serving NestJS

Steps to Reproduce:
1. Spin up new Nx Workspace
2. Generate Node app (NestJS) (app named foo)
3. Spin up app (ex: `npm run start foo`)
4. Navigate or refresh in Angular Console

Observe: The window goes blank and in the devtools you'll see these errors:
generate___nx-nestjs-app-breaks-angular

If you refresh a bit on the console you'll see `WebSockets request was expected`.

If you kill the NestJS app, the angular console will start working again.

mbriggs:
this is due to conflict with default nest debugging port. angular console will use a random port if it conflicts, but nest does not. 

resolves #569

### DIFF
--- a/apps/electron/src/main.ts
+++ b/apps/electron/src/main.ts
@@ -134,7 +134,7 @@ function createWindow() {
     mainWindow.show();
   });
 
-  getPort({ port: 7777 }).then((port: number) => {
+  getPort({ port: 7968 }).then((port: number) => {
     try {
       startServer(port, telemetry, store, mainWindow).then(() => {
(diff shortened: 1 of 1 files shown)


mbriggs:
The problem is that 7777 is open when angular-console boots up, it is nest from the generated app that wants that port when you serve it. Unless you are saying you want to fix this in the nest schematics

PR #801, opened 2019-08-19
Rework Analytics

mbriggs:


libs/server/src/lib/telemetry/user.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    if (settings && !settings.canCollectData) {
+      state = 'untracked';
+    }
+
+    const user = new User(id, state, isConnected);
+    return user;
+  }
+
+  onChange: OnChange = () => {};
```
mrmeku: Why leave this method in?

mbriggs: it is used here https://github.com/nrwl/angular-console/pull/801/files#diff-1911b3879856aa4f2cc5d325f594e7beR39

mbriggs: It is used in the google analytics sink to respond to logged in state changing.

CONTRIBUTING.md
```diff
@@ -21,7 +21,9 @@ After cloning the project run: `yarn`.
 
 After that, run `nps prepare.electron`. Every time you add or remove dependencies in electron/package.json, you will need to rerun `nps prepare.electron`.
 
-After this, run `nps dev.up` to start the dev environment. The application will start the process listening on port 4200. The development is done in the browser, but the server uses electron.
+After this, run `nps dev.up` to start the dev environment. The application will start the process listening on port 4200. The development is done in the browser, but the server uses electron. This is the most straight forward way of building new functionality.
```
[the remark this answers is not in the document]
mbriggs: I just added

> This is the most straight forward way of building new functionality.

but honestly, not married to it. My thinking was to call that out explicitly, since there is many ways to run it.

CONTRIBUTING.md
```diff
@@ -38,6 +40,24 @@ Cypress, which we use to run e2e tests, records the videos of the tests ran on C
 
 ## Building Electron App
 
+Before starting, be sure that you have run `nps prepare.electron` at least once (as previously mentioned).
+
+Building the electron app requires a code signing certificate to be set. You can read more [here](https://www.electron.build/code-signing).
```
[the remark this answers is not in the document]
mbriggs: we already have win and mac, the mac side at least verifies `CSC_LINK` and `CSC_PASSWORD`. The former would be easy enough to drop, but not sure about the implications.

CONTRIBUTING.md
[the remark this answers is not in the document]
mrmeku: This comment was targeted for line 26 of the readme. Sorry I misclicked

libs/server/src/lib/api/executable.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  get path() {
+    return this.isNvm ? `nvm exec ${this._path}` : this._path;
+  }
+
+  setPath(path: string) {
+    this._path = path;
+  }
+
+  setName(name: string) {
```
mrmeku: You already have a getter, why not just have a setting too?

mrmeku: I get this feeling someday someone is going to just try to set the property and be really consfused when nothing happens

libs/server/src/lib/api/executable.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    const workspace = workspaceName(type, cwd);
+    const statusCalculator = createDetailedStatusCalculator(cwd, cmds);
+
+    const id = `${this.name} ${cmds.join(' ')} ${commandRunIndex++}`;
+    let command = `${this.path} ${cmds.join(' ')}`;
+
+    // We currently don't support the windows implementation of NVM.
+    if (this.isNvm) {
```
mrmeku: You aren't doing a windows check here so the comment above doesn't make sense

libs/server/src/lib/telemetry/record.ts
```diff
@@ -0,0 +1,24 @@
...
+  | 'LoggedOut'
+  | 'StoppedTracking'
+  | 'StartedTracking'
+  | 'ScreenViewed'
+  | 'CommandRun'
+  | 'ExceptionOccurred'
+  | 'FeatureUsed';
+
+export function isTelemetryRecord(evt: string): evt is TelemetryType {
+  return [
```
mrmeku: Use a set rather than an array

libs/server/src/lib/telemetry/telemetry-client.spec.ts
```diff
@@ -0,0 +1,89 @@
+// import { TelemetryClient } from './telemetry-client';
```
mrmeku: Oh where oh where did all these tests go...

libs/server/src/lib/telemetry/user.ts
mrmeku: If you are implementing an interface or abstract class could you annotate this class with that information?

CONTRIBUTING.md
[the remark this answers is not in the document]
mbriggs: I may be missing something, that boots electron for me but doesn't open any windows?

libs/server/src/lib/api/executable.ts
mbriggs: you are totally right

libs/server/src/lib/api/executable.ts
mbriggs: Yeah, I preserved that from the existing code, I think the intention is to explain why that check isn't there. Can remove if you like

libs/server/src/lib/telemetry/record.ts
mbriggs: 👍

libs/server/src/lib/telemetry/telemetry-client.spec.ts
mbriggs: oof, should have deleted that >.> telemetry client was a remnant of the message bus based approach I backed out of

libs/server/src/lib/telemetry/user.ts
mbriggs: Changed the way that user works. The idea of `User` is that it handles the state machine around user state relating to telemetry. I had the google sink reading from it directly, I think the better idea is that User is something telemetry uses, and turns into another type of record. That removes a collaboration, and it puts that in the central mechanism of telemetry writing records, think its a better design :)

mbriggs:
Just a heads up, this is (manually) untested. Will try to find some time tomorrow to go through everything and make sure it all works <3 In theory, it should be mergable

libs/server/src/lib/api/executable.ts
mrmeku: Meh, keep it

PR #825, opened 2019-09-04
initial generation of vscode-ui

mrmeku:


package.json
```diff
@@ -81,11 +81,20 @@
     ]
   },
   "dependencies": {
+    "@angular/animations": "^8.0.0",
```
nixallover: Do we need these Angular packages in both the dependencies and devDependencies now?

mrmeku: Absolutely not! Good catch

PR #831, opened 2019-09-11
feat: use vscode settings for configuration

mbriggs:
- use vscode settings for things we want to expose to users for configuration
 - keep using global state for things we dont
 - unfortunately, this means we need to store information about whether migration has been performed
 - currently, it is using shared memory to update telemetry when that setting changes, when that should be exposed on the server. Looked into adding a graphql client to the plugin side of things, but didn't seem worth it in the context of this work


apps/vscode/src/app/workspace-tree/workspace-tree-item.ts
```diff
@@ -27,7 +27,6 @@ export type WorkspaceRouteTitle =
   | 'Xi18n'
   | 'Dep-Graph'
   | 'Connect'
-  | 'Settings'
```
mrmeku: Rather than removing this UI shortcut to settings, could we instead use this shortcut to open angular-console's vscode settings pane? I think in the short term it'll help teach our users that we have migrated to using the native settings apis

PR #859, opened 2019-10-21
feat(vscode): autocomplete

nixallover:
Created autocomplete input component for VS Code

- The autocomplete component contains an input as well, which is different than most implementations. I can refactor it to attach to another component, now that the initial version is working.
- Zack suggested I use nested form controls instead of the current strategy I'm using in select, input, and checkbox, so I did that here. If this seems like a good configuration, I'll update the other inputs to use it too.
- I decided not to use the CDK overlay as it was causing some problems, but I do still have a version with it available if it should be used.
- I'm planning to work on tests next.

### DIFF
--- a/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
+++ b/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
@@ -12,8 +12,6 @@
       class="text-input"
       type="text"
       role="textbox"
-      [required]="field.required"
-      [attr.list]="field.enum ? field.name + '_enum' : undefined"
     />
 
     <ng-container *ngIf="(isFocused$ | async)">
--- a/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
+++ b/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
@@ -1,3 +1,38 @@
-<datalist id="{{ field.name }}_enum">
-  <option *ngFor="let value of field.enum" [value]="value"></option>
-</datalist>
+<div class="description">
+  {{ field.description }}
+</div>
+
+<div class="modified-indicator"></div>
+
+<div class="control">
+  <div class="autocomplete-container">
+    <input
+      #textInput
+      [formControl]="control"
+      class="text-input"
+      type="text"
+      role="textbox"
+      [required]="field.required"
+      [attr.list]="field.enum ? field.name + '_enum' : undefined"
+    />
+
+    <ng-container *ngIf="(isFocused$ | async)">
+      <ng-container *ngIf="(visibleOptions | async) as result">
+        <div class="autocomplete-panel" role="listbox">
+          <div
+            *ngFor="let option of result; let i = index"
+            (click)="optionSelected(option)"
+            class="option"
+            role="option"
+            [class.selected]="(selectedOption$ | async) === i"
+          >
+            {{ option }}
+          </div>
+          <div *ngIf="!result.length" class="option no-result">
+            No matches for "{{ control.value }}"
+          </div>
+        </div>
+      </ng-container>
+    </ng-container>
+  </div>
+</div>
--- a/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
+++ b/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
@@ -1,16 +1,225 @@
-import { Component, Input, ChangeDetectionStrategy } from '@angular/core';
+import {
+  Component,
+  OnDestroy,
+  Input,
+  SimpleChanges,
+  OnChanges,
+  ViewChild,
+  ElementRef,
+  OnInit,
+  ViewEncapsulation
+} from '@angular/core';
+import {
+  ControlValueAccessor,
+  NG_VALUE_ACCESSOR,
+  FormControl
+} from '@angular/forms';
+import {
+  map,
+  tap,
+  takeUntil,
+  switchMap,
+  startWith,
+  scan,
+  debounceTime,
+  filter,
+  shareReplay
+} from 'rxjs/operators';
+import {
+  Subject,
+  BehaviorSubject,
+  Observable,
+  fromEvent,
+  merge,
+  of
+} from 'rxjs';
 import { Schema } from '@angular-console/schema';
 
-/**
- * Not built yet - using a datalist temporarily to loosely mimic functionality
- */
+enum AutocompleteNavKeys {
+  Enter = 'Enter',
+  ArrowUp = 'ArrowUp',
+  ArrowDown = 'ArrowDown'
+}
 
 @Component({
   selector: 'angular-console-autocomplete',
   templateUrl: './autocomplete.component.html',
   styleUrls: ['./autocomplete.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush
+  encapsulation: ViewEncapsulation.None,
+  providers: [
+    {
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: AutocompleteComponent,
+      multi: true
+    }
+  ]
 })
-export class AutocompleteComponent {
+export class AutocompleteComponent
+  implements OnInit, OnDestroy, OnChanges, ControlValueAccessor {
   @Input() field: Schema;
+
+  @Input() options: string[] = [];
+  private readonly _options$ = new BehaviorSubject<string[]>(this.options);
+  visibleOptions: Observable<string[]>;
+  selectedOption$: Observable<number | null>;
+
+  @ViewChild('textInput', { static: true }) textInput: ElementRef;
+  control: FormControl;
+  isFocused$: Observable<boolean>;
+
+  private readonly _destroying$ = new Subject<void>();
+
+  constructor(private readonly _elementRef: ElementRef) {}
+
+  writeValue(value: string) {
+    if (this.control) {
+      this.control.setValue(value);
+    } else {
+      this.control = new FormControl(value);
+      this.visibleOptions = this._options$.pipe(
+        switchMap(options =>
+          this.control.valueChanges.pipe(
+            startWith(this.control.value),
+            map(formValue =>
+              options.filter(option =>
+                option.toLowerCase().includes(formValue.toLowerCase())
+              )
+            )
+          )
+        )
+      );
+    }
+  }
+
+  ngOnInit() {
+    this.isFocused$ = merge(
+      fromEvent(this._elementRef.nativeElement, 'focusin').pipe(
+        map(() => 'focus in')
+      ),
+      fromEvent(this._elementRef.nativeElement, 'focusout').pipe(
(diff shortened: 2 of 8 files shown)


libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
```diff
@@ -1,3 +1,38 @@
...
+<div class="control">
+  <div class="autocomplete-container">
+    <input
+      #textInput
+      [formControl]="control"
+      class="text-input"
+      type="text"
+      role="textbox"
+      [required]="field.required"
+      [attr.list]="field.enum ? field.name + '_enum' : undefined"
```
mrmeku: Do we need this binding? We aren't using data lists anymore and so far as I've read this attribute is meant to hook up an input to a datalist

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
```diff
@@ -1,3 +1,38 @@
...
+
+<div class="control">
+  <div class="autocomplete-container">
+    <input
+      #textInput
+      [formControl]="control"
+      class="text-input"
+      type="text"
+      role="textbox"
+      [required]="field.required"
```
mrmeku: Do we need this binding? What does it give us in terms of UX? If it doesn't add value remove it since removing as many bindings as we can improves our time to initial render

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -1,16 +1,225 @@
...
+  @Input() options: string[] = [];
+  private readonly _options$ = new BehaviorSubject<string[]>(this.options);
+  visibleOptions: Observable<string[]>;
+  selectedOption$: Observable<number | null>;
+
+  @ViewChild('textInput', { static: true }) textInput: ElementRef;
+  control: FormControl;
+  isFocused$: Observable<boolean>;
+
+  private readonly _destroying$ = new Subject<void>();
```
mrmeku: The rest of the code base doesn't use _ prefixing for private variables nor do we use the $ suffix for subjects. Consider sticking to our current conventions for consistency

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -1,16 +1,225 @@
...
+
+  writeValue(value: string) {
+    if (this.control) {
+      this.control.setValue(value);
+    } else {
+      this.control = new FormControl(value);
+      this.visibleOptions = this._options$.pipe(
+        switchMap(options =>
+          this.control.valueChanges.pipe(
+            startWith(this.control.value),
```
mrmeku: James Henry filed a bug once requesting that when an autocomplete is opened initially (even if it already has a value) we start by showing the full list of options

Consider the case in which a configuration specifies that a particular enum value be used. When the user clicks into the autocomplete they initially only see the pre-defined options and have to clear the input to see what other options exist. I think it adds value to show all options until the user starts typing and actively filtering the results

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -1,16 +1,225 @@
...
+          )
+        )
+      );
+    }
+  }
+
+  ngOnInit() {
+    this.isFocused$ = merge(
+      fromEvent(this._elementRef.nativeElement, 'focusin').pipe(
+        map(() => 'focus in')
```
mrmeku: This would probably be the appropriate place to reset the list of filtered options to show the entire list

mrmeku: Also maybe just map the focusin event to true and the focus out event to false. Then you can remove the map below

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -1,16 +1,225 @@
...
+      this._options$.next(this.options);
+    }
+  }
+
+  /**
+   * Check if the input key is one relevant to autocomplete panel navigation
+   */
+  isNavigationKey = (key: KeyboardEvent['key']): boolean => {
+    return (
+      key === AutocompleteNavKeys.ArrowUp ||
```
mrmeku: It would be slightly more efficient and idiomatic to define a module level constant of the et of AutocompleteNavKeys this function cares about

libs/vscode-ui/components/src/lib/field/field.component.html
```diff
@@ -4,22 +4,28 @@
...
-    [(value)]="value"
-  ></angular-console-checkbox>
 
-  <angular-console-select
-    *ngSwitchCase="'enum'"
-    [field]="field"
-    [(value)]="value"
-  ></angular-console-select>
+<angular-console-autocomplete
+  *ngIf="field.type === 'array'"
```
mrmeku: I think our logic here is that we want to use an angular-console-select if the number of enum values is 10 or less and an autocomplete if 11 or more. We do not currently map to an `array` data type. 

We could definitely augment the server to do so. The server could check if the number of selectable values is 10 or less and conditionally set the type that way. 

What do you think ?

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
nixallover: No we don't, I'll remove it.

libs/vscode-ui/components/src/lib/field/field.component.html
nixallover: Oh I see. I was thinking that there might be some scenarios where we can provide a list of options, but the user could put in a value outside of that list. Is that not the case, or is there a way we could account for that?

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
nixallover: `list` binding removed

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
nixallover: `required` binding removed

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
nixallover: Renamed to `destroying`

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
nixallover: I updated this to start with ''

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
nixallover: Done

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
nixallover: Done

libs/vscode-ui/components/src/lib/field/field.component.html
mrmeku: I don't think there's any need to account for that currently

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -1,16 +1,222 @@
...
-/**
- * Not built yet - using a datalist temporarily to loosely mimic functionality
- */
+export enum AutocompleteNavKeys {
+  Enter = 'Enter',
+  ArrowUp = 'ArrowUp',
+  ArrowDown = 'ArrowDown'
+}
+
+export const AUTOCOMPLETE_NAV_KEYS = Object.values(AutocompleteNavKeys);
```
mrmeku: make this into an es6 Set

expost const AUTOCOMPLETE_NAV_KEYS = new Set('Enter', 'ArrowUp', 'ArrowDown');

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -1,16 +1,222 @@
...
+  let modifier: -1 | 0 | 1 = 0;
+
+  if (key === AutocompleteNavKeys.ArrowDown) {
+    modifier++;
+  }
+  if (key === AutocompleteNavKeys.ArrowUp) {
+    modifier--;
+  }
+
+  const temp = currentIndex + modifier;
```
mrmeku: nit: consider calling this something like `selectedIndex` for code clarity.

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -1,16 +1,222 @@
...
+ * side of the array (to loop around the list).
+ * @param key {AutocompleteNavKeys} KeyboardEvent.key value
+ * @param currentIndex {number} index before user input
+ * @param max {number} total number of options
+ * @returns new option index
+ */
+function updatedOptionIndex(
+  key: AutocompleteNavKeys,
+  currentIndex: number,
+  max: number
```
mrmeku: nit: Consider naming this something along the lines of `numOptions` for code clarity

libs/vscode-ui/components/src/lib/field/field.component.ts
```diff
@@ -23,9 +29,31 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
...
 })
-export class FieldComponent implements ControlValueAccessor {
+export class FieldComponent implements ControlValueAccessor, OnDestroy {
   @Input() field: Schema;
   _value: string;
+  valueChangeSub: Subscription;
+
+  control = new FormControl('');
+
+  exampleOptions = [
```
mrmeku: The options will be within `field.enum`

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
nixallover: done

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
nixallover: done

libs/vscode-ui/components/src/lib/field/field.component.ts
nixallover: Fixed

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
nixallover: Good call!

PR #866, opened 2019-10-29
chore(vscode-ui): remove Connect and Dep-Graph routes from vs code ex…

jaysoo:
…tension


dbrody:
Why was this removed? Is there a way to use the Dep-Graph in the current version?

PR #878, opened 2019-11-08
chore: Remove dead electron code

mrmeku:


jaysoo:
Looks like some build failures, and `nx.json` being out of sync. Thanks for doing this!

Issue #880, opened 2019-11-11
Deploy command is invalid


smarth55:
After the update to the VSCode plugin, the deploy command that can be selected from the Angular Console > Navigation does not run. I can see that when I try to run it, the terminal outputs `Executing task: ng deploy my-app <`
but errors with `The specified command ("deploy") is invalid.`

I can, however, select the run command and the choose the deploy command from that menu and get `Executing task: ng run my-app:deploy <` which does work.

Just a minor inconvenience as long as it still deploys, but still confusing.

isaacplmann:
Good catch.  I can reproduce that as well.  The `ng deploy my-app` command should be `ng run my-app:deploy`.

The short-hand version works for `serve`, `build` and `test`, but not for  tasks that haven't been hard coded by the Angular CLI.

mrmeku:
Ah that makes sense. Deploy is hard coded in later version of the cli but not in some earlier ones. We can change things so that deploy doesn't use the short hand syntax so that we can support older versions of the CLI

jaysoo:
Thanks for the bug report. This will be fixed in the next patch release.

PR #888, opened 2019-11-14
fix(vscode): use longer version of `[prog] run [project]:[command}` t…

jaysoo:
…o improve support for different CLI versions.


jaysoo:
Closes #880

PR #889, opened 2019-11-14
fix(vscode): Allow for multiple tasks of same type

mrmeku:
Fixes https://github.com/nrwl/angular-console/issues/883


mrmeku:
@jaysoo For context, vscode kinda sorta uses a tasks name to tell if its a duplicate or not. Now I'm just naming the tasks the same as their command line arguments since that seems like a good identifier

PR #890, opened 2019-11-14
fix(windows): Properly escape spaces in powershell with backticks

mrmeku:


mrmeku:
Fixes https://github.com/nrwl/angular-console/issues/857 and https://github.com/nrwl/angular-console/issues/821

PR #892, opened 2019-11-15
fix(vscode-ui): Correctly populate defaults

mrmeku:
Fixes parsing defaults from angular json
Fixes setting defaults from schematic fields (many were appearing as false)
Fixes setting default for select menus


libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.ts
```diff
@@ -294,7 +294,7 @@ export class TaskExecutionFormComponent implements OnInit, AfterViewChecked {
     const defaultValues: { [key: string]: string } = {};
     architect.schema.forEach(field => {
       defaultValues[field.name] =
-        field.defaultValue || field.type === 'boolean' ? 'false' : '';
+        field.defaultValue || (field.type === 'boolean' ? 'false' : '');
```
mrmeku: Stupid order or operations

PR #909, opened 2019-11-25
fix(vscode): fix opt-out event for data collection

jaysoo:
- Record "opt out" events before we stop collecting usage data.
- Reads from vscode configuration instead of our store since it doesn't appear we are updating the store anywhere.

### DIFF
--- a/apps/vscode/src/app/telemetry.ts
+++ b/apps/vscode/src/app/telemetry.ts
@@ -11,6 +11,10 @@ export function getTelemetry() {
   return telemetry;
 }
 
+const enableTelemetry = 'enableTelemetry';
+const configurationSection = VSCodeStorage.configurationSection;
+const telemetrySetting = `${configurationSection}.${enableTelemetry}`;
+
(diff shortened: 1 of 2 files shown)


apps/vscode/src/app/telemetry.ts
```diff
@@ -28,10 +28,13 @@ export function initTelemetry(context: ExtensionContext, store: Store) {
   const enableTelemetry = 'enableTelemetry';
   const configurationSection = VSCodeStorage.configurationSection;
   const telemetrySetting = `${configurationSection}.${enableTelemetry}`;
+  const configuration = workspace.getConfiguration(
+    VSCodeStorage.configurationSection
+  );
 
   disposer = workspace.onDidChangeConfiguration(e => {
     if (e.affectsConfiguration(telemetrySetting)) {
-      if (store.get(enableTelemetry, true)) {
```
mrmeku: This is the line that has tracking default to true. Remove it will make telemetry opt in rather than opt out

PR #913, opened 2019-11-26
feat(server): Use angular-cli's built in parser

### LINKED ISSUE #908: [schematic] generate
If the property in schema is set as visible: false

https://github.com/ngrx/platform/blob/c9d1f23c6770b2470b97ce6a8683dc754bad4505/modules/schematics/src/feature/schema.json#L11

there is no way to set the property through UI

However, it is nice to set things like path or UI should pick up the path from explorer ui.

mrmeku:
Fixes #908


mrmeku:
Fixes https://github.com/nrwl/angular-console/issues/908

libs/schema/src/index.ts
```diff
@@ -1,15 +1,7 @@
...
-  completion?: string;
-  deprecated?: string;
-  required: boolean;
-  positional: boolean;
-  enum?: string[];
-}
+import { Option } from '@angular/cli/models/interface';
+
+// tslint:disable-next-line: no-empty-interface
+export interface Schema extends Option {}
```
jaysoo: Is the `Schema` just an alias now? Like `type Schema = Option`?

mrmeku: Bingo. 

In a followup I want to remove the `Schema` interface all together and just use `Option` directly. I made an attempt to do it as part of the same change but it produced a lot of noise

jaysoo: Ah cool :)

PR #914, opened 2019-11-26
chore: Code cleanliness

mrmeku:
Rename Schema interface to Option (exported from angular CLI)

Remove many utility methods that are not called

### DIFF
--- a/libs/server/src/lib/utils/read-projects.ts
+++ b/libs/server/src/lib/utils/read-projects.ts
@@ -1,4 +1,10 @@
-import { Architect, Project, Schema } from '@angular-console/schema';
+import {
+  Architect,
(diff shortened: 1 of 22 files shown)


libs/server/src/lib/utils/read-projects.ts
```diff
@@ -9,43 +15,31 @@ import {
...
-          architectDef.configurations,
-          name
-        )
+        defaultValues: readDefaultValues(architectDef.configurations, name)
       }))
     : [];
 
   return {
-    schema: [],
-    options,
```
jaysoo: Seems like the old `options` value was never used before right? So replacing it has no effect on behavior.

PR #926, opened 2020-01-22
feat(vscode): Support workspace.json for non Angular workspaces

mrmeku:


igor-radivojevic-ct:
Hey @jaysoo,
I can see that this change is approved and merged, but I can not see how to use it.
Is it released or it is planned to be released soon?

Issue #923, opened 2020-01-02
Generate navigation menu fails to list available commands in version 10.0.0


ctaleck:
Angular console version _10.0.0_ disabled the **Generate** menu item and does not display the command palette for my project. When I downgrade to version _9.1.1_, the command works by displaying the list of commands. 

I have the `angular.json` and `nx.json` files in my workspace. I do not have the new `workspace.json` file.

StevenWouters:
I have the same problem. 

stack trace: TypeError: Cannot convert undefined or null to object	at Function.entries (<anonymous>)	at readCollectionSchematics (c:\foldername\nrwl.angular-console-10.0.0\main.js:167374:12)	at readCollection (c:\foldername\nrwl.angular-console-10.0.0\main.js:167363:16)

KingDarBoja:
I get the same issue, the **generate** option doesn't display tje cpmmand palette.

More info:
```
Version: 1.41.1 (user setup)
Commit: 26076a4de974ead31f97692a0d32f90d735645c0
Date: 2019-12-18T14:58:56.166Z
Electron: 6.1.5
Chrome: 76.0.3809.146
Node.js: 12.4.0
V8: 7.6.303.31-electron.0
OS: Windows_NT x64 10.0.18363
```

AlbertoD94:
Same

alejandrogkl:
Same here, when downgrade to 9.1.1 the generate command palette works

Jejuni:
I've got the same error :
`[[object Object]]Cannot convert undefined or null to object`

`Cannot convert undefined or null to object
	at Function.entries (<anonymous>)
	at readCollectionSchematics (c:\Users\Jejun\.vscode\extensions\nrwl.angular-console-10.0.0\main.js:167374:12)
	at readCollection(...)`

I've narrowed it down to the suite of Telerik products we have installed.
After further inspection some packages in the node_modules/@progress folder cause the issues.
Among them was, for instance, `kendo-angular-l10n`.

This package, unlike most others, did not have a `schematics` folder. However, in the included package.json there were 2 entries referencing the schematics: 
`(...)
"schematics": "./schematics/collection.json",
  "files": [
    "dist",
    "schematics"
  ]
(...)`

Removing these entries fixed the problem.
However, after reinstalling or updating the package.json will be regenerated and the problem starts appearing again.
Could this be fixed so a non-existing schematics entry/folder doesn't cause the crash?

alejandrogkl:
[...]

PR #971, opened 2020-08-28
Handle missing schematics json file

isaacplmann:
The generate command becomes unresponsive when there is an npm package with a `package.json` file that has a `schematics` property that does not point to a valid file.

This PR will skip the invalid `package.json` file instead of hanging.


ctaleck:
Does this fix #923 also?

isaacplmann:
It should.  Reopen that issue if it is still broken

Issue #974, opened 2020-09-05
Feature: Open NX generate UI from the VSCode file explorer context menu and use context information to pre-fill path and project


glenngr:
Hi. I'm looking for feedback on an NX Console feature idea.

I've forked the repo and played around abit with the idea of being able to right click on a folder and selecting generate (ui).

When you select a schematic, the path and project fields are automatically filled out based on where you clicked in the context menu.

Here's what I have done: https://github.com/nrwl/nx-console/compare/master...glenngr:feat/invoke-nx-console-from-file-explorer-context-menu

Are you guys interested in this? Should I open a PR? 

I'd like feedback on how I've implemented it. I'm sure there are better ways to do certain things.

isaacplmann:
Definitely open a PR.  I think your code changes all make sense.  I have some UX related concerns:

1. There should be a setting to disable the options from showing in the context menu (I haven't researched how to do this)
2. It would be good to have the context menu item say "Nx: generate (ui)" instead of just "generate (ui)".
3. Also, make sure that the context menu works correctly in a plain Angular CLI workspace as well

Let me know if you want some help on this.  I'm excited to get this feature added.

glenngr:
Thanks. I agree with all your suggestions.
I'll look into them, open a PR and ask for help when I need it.

PR #977, opened 2020-09-10
feat: show generate ui from explorer context menu

glenngr:
nx generate ui can be shown by right clicking in explorer context menu
the path you click on will be used as a value for the "path" option
the "project" option will also be updated based on the path 

See discussion in #974, issues listed there has been solved. 

I have only tested on windows so far. Should we make sure it works in mac and linux as well?

Any more changes or suggestions?

### DIFF
--- a/apps/vscode/src/app/cli-task/cli-task-commands.ts
+++ b/apps/vscode/src/app/cli-task/cli-task-commands.ts
@@ -1,4 +1,4 @@
-import { commands, ExtensionContext, window } from 'vscode';
+import { commands, ExtensionContext, window, Uri } from 'vscode';
 
 import { selectSchematic } from '../select-schematic';
 import { verifyWorkspace } from '../verify-workspace/verify-workspace';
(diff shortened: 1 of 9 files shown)


isaacplmann:
I'll test on mac.

isaacplmann:
This looks great.  After trying this out with an assortment of schematics, I think you should also auto-populate a `directory` field in a schematic with the same value as `path`.  Once that's in place, I'll merge it in.

glenngr:
Just set "directory" to the same value as "path", like in my latest force-push update?

glenngr:
Should the README or other documentation be updated as well? Maybe how to disable the menu item through the settings?

isaacplmann:
I'm sorry, I made a bad suggestion.  Apparently all the `directory` fields don't work the same way as the `path` fields.  The `directory` fields want a path relative to the project root, not the workspace root.

I think we should revert those changes and potentially revisit that in a separate PR.  There are a lot of weird edge cases to the directory field.

So if you switch it back to just populating `path` and `project`, I'll merge this in.

glenngr:
Ok, no problem. I have reverted the change now.

isaacplmann:
Thanks! I'll test and merge this on Monday

isaacplmann:
Thanks!  This is a great feature

glenngr:
@isaacplmann when will the new version of the vscode plugin be released?

Issue #987, opened 2020-11-03
Feature: add field as argument even when it is the default value


srlee309:
At the moment, if you don't change a field from the default value it needs to be confirmed in the angular cli when running the command.
https://github.com/nrwl/nx-console/blob/master/libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.ts#L345

Can it be changed so that it adds the field as an argument even when it is the default value. This would stop the angular cli asking you to confirm that you want to use the default value. If you want to keep this confirmation request, can be restricted so that it only happens for the required fields. This would mean that if the field is the default value and not a required field it would be added as an argument in the command and the user would not need to confirm it.

sandikbarr:
Fixed in #1008 and deployed today in v17.0.0

PR #988, opened 2020-11-11
fix: add default field as argument

ImCesar:
Pass the default field as an argument so the angular CLI does not prompt you for the default again

### DIFF
--- a/libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.ts
+++ b/libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.ts
@@ -252,7 +252,11 @@ export class TaskExecutionFormComponent implements OnInit, AfterViewChecked {
       }
       taskExecForm.addControl(
         schema.name,
-        new FormControl((architect.contextValues && architect.contextValues[schema.name]) || defaultValues[schema.name], validators)
+        new FormControl(
+          (architect.contextValues && architect.contextValues[schema.name]) ||
+            defaultValues[schema.name],
+          validators
+        )
       );
     });
 
@@ -277,7 +281,6 @@ export class TaskExecutionFormComponent implements OnInit, AfterViewChecked {
       top: 0
     });
   }
-
 
   private getDefaultValuesForConfiguration(
     architect: TaskExecutionSchema,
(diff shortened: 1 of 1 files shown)


ImCesar:
This commit fixes #987

srlee309:
Will this include a value like this: "--exclude=" when the schematic option is something like the below, i.e. default empty array.
```
"exclude": {
      "type": "array",
      "description": "Files to exclude from linting.",
      "default": [],
      "items": {
        "type": "string"
      }
}
```

s4m0r4m4:
I was just looking into fixing this myself! Glad to see somebody got to it first. @isaacplmann any chance you could take a quick look at this easy fix?

s4m0r4m4:
@sandikbarr - I see you've been pretty active in this repo, would you be able to take a look at this one? See my comment above for more discussion. I'd like to see this resolved and am willing to contribute a PR if anything needs to be changed/updated.

sandikbarr:
@s4m0r4m4 yes, I will take a look at this one too. It looks like a good update. We were talking about the x-prompt fields and defaults recently as well. I am new to taking on Nx Console maintenance, so I will talk this over with folks and get back to y'all.

sandikbarr:
@ImCesar @s4m0r4m4 I put in #1008 which I think should get what you were after here. The gist is that we'll run commands with --no-interactive which then won't prompt for values when there is a default. We don't want to include the defaults in the command because the schematic or generator should read the defaults from the schema instead. This keeps us from duplicating too much functionality in this extension as well as keeps the commands shorter without unnecessarily including defaults. Please let me know your thoughts.

s4m0r4m4:
Thanks @sandikbarr , that PR looks good to me. I do agree with the one comment from srlee about running it with `defaults` flag instead, that makes more sense to me (I'd still want it to prompt me if there was a field requiring user input that did _not_ have a default value). I'm excited to see the project dropdown fix in there too!

sandikbarr:
> I do agree with the one comment from srlee about running it with `defaults` flag instead, that makes more sense to me (I'd still want it to prompt me if there was a field requiring user input that did _not_ have a default value).

Yeah, the defaults flag does seem more appropriate there. I think I would also want it to go ahead and prompt if for some reason the x-prompt default hasn't been read correctly or something wasn't quite supported right so that we still have that fall back. I'll make sure I'm not missing anything with that default flag and probably make that change.

#1008 makes x-prompts that do not have defaults required fields. The dryRun will go ahead when the form is not valid, but the Run button is disabled if the form does not pass validation. So when running the command, the intention is it shouldn't hit the x-prompt when it doesn't have a default because a value had to be provided.

> I'm excited to see the project dropdown fix in there too!

Yes, so glad to have that project dropdown fixed.

Tysm for your feedback!

sandikbarr:
resolved with #1008

PR #993, opened 2020-12-23
support for Nx 11

sandikbarr:
Updates Nx Console with support for Nx 11 configuration.


vsavkin:
I took a look. I think it might be easier if you normalize the data when you read it. For instance this is what we have when read the workspace config in nx itself.


export function toNewFormat(w: any) {
  Object.values(w.projects || {}).forEach((project: any) => {
    if (project.architect) {
      renameProperty(project, 'architect', 'targets');
    }
    if (project.schematics) {
      renameProperty(project, 'schematics', 'generators');
    }
    Object.values(project.targets || {}).forEach((target: any) => {
      if (target.builder) {
        renameProperty(target, 'builder', 'executor');
      }
    });
  });

  if (w.schematics) {
    renameProperty(w, 'schematics', 'generators');
  }
  return w;
}

If you invoke a function like this in, say, verify-workspace.ts, the rest will be a simple rename of architect to taget or schematict to generator. No if statement will be added. This is even more important as soon we will split the workspace.json into multiple files and having a single place to stick them all together together will make a lot of things simpler.

Same for loading schema files. If you normliaze the data when you read it, you won't have to deal with two options every single step along the way.

The normalization approach can get tricky if you want to mutate those files and you want to preserve the original names, but that's not the case for you. Console never writes, only reads.

If you want to make it work quickly, you can actually go in the other direction first and normalize to the old format. If you do that, the only places that will have to change are: readSchematicCollections, readBuilderSchema and verifyWorkspace. Everything else should rename the same because no types will change.

But, as I mentioned, we can do it in January.

PR #996, opened 2021-01-04
convert to legacy format to support Nx 11

sandikbarr:
converts workspace, package.json, generators, and executors to Legacy formats to support Nx11

### DIFF
--- a/libs/server/src/lib/utils/read-projects.ts
+++ b/libs/server/src/lib/utils/read-projects.ts
@@ -75,7 +75,7 @@ export async function readBuilderSchema(
     path.join(npmPackage, 'package.json'),
     path.join(basedir, 'node_modules')
   );
-  const b = packageJson.json.builders;
+  const b = toLegacyFormat(packageJson.json).builders;
   const buildersPath = b.startsWith('.') ? b : `./${b}`;
   const buildersJson = readAndCacheJsonFile(
     buildersPath,
--- a/libs/server/src/lib/utils/read-projects.ts
+++ b/libs/server/src/lib/utils/read-projects.ts
@@ -81,7 +81,8 @@ export async function readBuilderSchema(
     path.dirname(packageJson.path)
   );
 
-  const builderDef = (buildersJson.json.builders || buildersJson.json.executors)[builderName];
+  const builderDef = (buildersJson.json.builders ||
+    buildersJson.json.executors)[builderName];
   const builderSchema = readAndCacheJsonFile(
     builderDef.schema,
     path.dirname(buildersJson.path)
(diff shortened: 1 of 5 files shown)


libs/server/src/lib/utils/utils.ts
```diff
@@ -290,3 +290,38 @@ export function getPrimitiveValue(value: any): string | undefined {
     return undefined;
   }
 }
+
+function renameProperty(obj: any, from: string, to: string) {
```
vsavkin, line 294: We do it in this way in the Nx repo itself cause we need to preserve the order of properties (cause we are serializing the json back to the file). You don't do the serialization, so the order should not matter. So you can do the rename as follows:

```
obj[to] = obj[from]
delete obj[from]
```

libs/server/src/lib/utils/read-projects.ts
```diff
@@ -74,14 +75,14 @@ export async function readBuilderSchema(
     path.join(npmPackage, 'package.json'),
     path.join(basedir, 'node_modules')
   );
-  const b = packageJson.json.builders;
+  const b = toLegacyFormat(packageJson.json).builders;
```
vsavkin, line 78: toLegacyFormat only updates angular.json/workspace.json. You probably either have a similar one to package.json that renames generators into schematics and executors into builders or do it on the spot here.

libs/server/src/lib/utils/read-projects.ts
```diff
@@ -74,14 +75,14 @@ export async function readBuilderSchema(
...
   const buildersPath = b.startsWith('.') ? b : `./${b}`;
   const buildersJson = readAndCacheJsonFile(
     buildersPath,
     path.dirname(packageJson.path)
   );
 
-  const builderDef = buildersJson.json.builders[builderName];
+  const builderDef = toLegacyFormat(buildersJson.json).builders[builderName];
```
vsavkin, line 85: See the comment above. toLegacyFormat only updates workspace.json/angular.json. You need to a different toLegacyFormat here.

vsavkin (changes requested):
You need to have 4 normalization functions instead of one. You have one for workspace.json, but you also need one for package.json, generators.json and executors.json. The missing three should be very small (3 lines of code each). So you don't really have to have them as functions, you can also inline them.

libs/server/src/lib/utils/utils.ts
```diff
@@ -290,3 +290,33 @@ export function getPrimitiveValue(value: any): string | undefined {
...
+  Object.keys(obj).forEach(k => {
```
vsavkin, line 295: you don't need to loop, right? you aren't using `k` for anything.

sandikbarr: gah, sorry yeah. will fix.

PR #999, opened 2021-01-19
feat: consider the new workspace generators path

rarmatei:
With Nx11, we need to also consider the tools/generators path when looking for schematic collections

### DIFF
--- a/libs/server/src/lib/utils/read-schematic-collections.ts
+++ b/libs/server/src/lib/utils/read-schematic-collections.ts
@@ -8,28 +8,45 @@ import {
   listOfUnnestedNpmPackages,
   normalizeSchema,
   readAndCacheJsonFile,
-  readAndParseJson
+  readAndParseJson,
+  toLegacyWorkspaceFormat
 } from './utils';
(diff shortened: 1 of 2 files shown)


libs/server/src/lib/utils/read-schematic-collections.ts
```diff
@@ -8,30 +8,47 @@ import {
...
   readAndCacheJsonFile,
-  readAndParseJson
+  readAndParseJson,
+  toLegacyWorkspaceFormat
 } from './utils';
-import { toLegacyWorkspaceFormat } from './utils';
 
 export async function readAllSchematicCollections(
-  workspaceJsonPath: string,
-  workspaceSchematicsPath: string
```
rarmatei, line 17: this function didn't seem to be used anywhere else, and it was an easier implementation if I generated the `workspaceSchematicsPath` in this function, rather than sending it in - so that's why I removed the extra param

but I can add it back if anybody thinks it's good to have 👍

sandikbarr (approved):
good update, pulling that checkAndRead out and removing the extra param seems like a good call. tysm!

PR #1006, opened 2021-02-02
fix: add support for "targets" alias in addition to default "architect"

miluoshi:
Angular CLI supports field "targets" as alias of "architect". Nx console doesn't support this alias, but we use it in our project.

See this change in Angular CLI's schema.json: https://github.com/angular/angular-cli/commit/e6f9ae98f5c3688b6efb9390956b0b60fbfaf228

Specifically, this snippet shows that the two fields are equal:

```jsonc
{
  "architect": {
    "type": "object",
    "additionalProperties": {
      "$ref": "#/definitions/project/definitions/target"
    }
  },
  "targets": {
    "type": "object",
    "additionalProperties": {
      "$ref": "#/definitions/project/definitions/target"
    }
  },
  // ...
  "anyOf": [
    {
      "required": ["architect"],
      "not": {
        "required": ["targets"]
      }
    },
    {
      "required": ["targets"],
      "not": {
        "required": ["architect"]
      }
    }
  ]
}
```

### DIFF
--- a/libs/server/src/lib/utils/read-projects.ts
+++ b/libs/server/src/lib/utils/read-projects.ts
@@ -20,7 +20,7 @@ export function readProjects(json: any): Project[] {
         name: key,
         root: value.root,
         projectType: value.projectType,
-        architect: readArchitect(key, value.architect)
+        architect: readArchitect(key, value.architect || value.targets)
       })
     )
     .sort((a, b) => a.root.localeCompare(b.root));
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@
     <img src="https://raw.githubusercontent.com/nrwl/nx-console/master/static/nx-console.png" width="500"/>
 </p>
 
-
 # The UI for Nx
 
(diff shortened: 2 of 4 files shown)


libs/server/src/lib/utils/read-projects.ts
```diff
@@ -20,7 +20,7 @@ export function readProjects(json: any): Project[] {
         name: key,
         root: value.root,
         projectType: value.projectType,
```
sandikbarr, line 22: Sorry if this is a dumb question 😬  Do you know what calls this `readProjects` function? Nx 11 supports targets too, and I added support for reading targets as architect in #996. However I skipped over this because I didn't see it used anywhere and thought it was dead code.

sandikbarr:
From v16 and on, Nx Console should be supporting `targets` in place of `architect`. If this is not the case for you, I'd like to follow up and look further into the specific issue it's having with your workspace.

PR #1008, opened 2021-02-04
validates form input and run command with --no-interactive

sandikbarr:
* Fixes project entries dropdown
* Makes x-prompt properties without a default required
* Shows form validation errors and disables Run if invalid
* Run all commands with --no-interactive, not just the dryRun

### DIFF
--- a/apps/vscode/src/app/workspace-tree/get-task-execution-schema.ts
+++ b/apps/vscode/src/app/workspace-tree/get-task-execution-schema.ts
@@ -121,13 +121,15 @@ export async function getTaskExecutionSchema(
             schematic.options.forEach(s => {
               // TODO: mixup between items and enum has been a source for recent bugs,
               //  util.ts normalizeSchema sets items from enum.
-              //  Should this getProjectEntries be over there?
               if (s.enum) {
                 return;
               }
 
-              if (s.name === 'project') {
-                s.items = cliTaskProvider
+              if (
+                s.name === 'project' ||
+                (s.$default && s.$default.$source === 'projectName')
+              ) {
+                s.enum = s.items = cliTaskProvider
                   .getProjectEntries()
                   .map(entry => entry[0])
                   .sort();
--- a/apps/vscode/src/app/workspace-tree/get-task-execution-schema.ts
+++ b/apps/vscode/src/app/workspace-tree/get-task-execution-schema.ts
@@ -119,12 +119,15 @@ export async function getTaskExecutionSchema(
             }
 
             schematic.options.forEach(s => {
+              // TODO: mixup between items and enum has been a source for recent bugs,
+              //  util.ts normalizeSchema sets items from enum.
+              //  Should this getProjectEntries be over there?
               if (s.enum) {
                 return;
               }
 
               if (s.name === 'project') {
-                s.enum = cliTaskProvider
+                s.items = cliTaskProvider
                   .getProjectEntries()
                   .map(entry => entry[0])
                   .sort();
--- a/apps/vscode/src/package.json
+++ b/apps/vscode/src/package.json
@@ -2,7 +2,7 @@
   "name": "angular-console",
   "displayName": "Nx Console",
   "description": "Nx Console for Visual Studio Code",
-  "version": "16.0.2",
+  "version": "17.0.0",
   "repository": {
     "type": "git",
     "url": "https://github.com/nrwl/nx-console.git"
--- a/libs/server/src/lib/utils/utils.ts
+++ b/libs/server/src/lib/utils/utils.ts
@@ -169,14 +169,15 @@ export async function normalizeSchema(
 
   const nxOptions = options.map(option => {
     const xPrompt: XPrompt = s.properties[option.name]['x-prompt'];
-    const defaultValue = projectDefaults && projectDefaults[option.name];
+    const workspaceDefault = projectDefaults && projectDefaults[option.name];
     const $default = s.properties[option.name].$default;
 
     const nxOption: Option = {
       ...option,
       component: getFieldType(option, xPrompt, $default && $default.$source),
       required: isFieldRequired(requiredFields, option, xPrompt, $default),
-      ...(defaultValue && {default: defaultValue}),
+      ...(workspaceDefault && {default: workspaceDefault}),
+      $default,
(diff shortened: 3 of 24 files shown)


apps/vscode/src/app/workspace-tree/get-task-execution-schema.ts
```diff
@@ -119,12 +119,15 @@ export async function getTaskExecutionSchema(
             }
 
             schematic.options.forEach(s => {
+              // TODO: mixup between items and enum has been a source for recent bugs,
+              //  util.ts normalizeSchema sets items from enum.
+              //  Should this getProjectEntries be over there?
```
sandikbarr, line 124: This fixes the projects dropdown, but the mixup between items and enum is still a bit confusing.

libs/server/src/lib/utils/utils.ts
```diff
@@ -169,14 +169,17 @@ export async function normalizeSchema(
```
sandikbarr, line 169: Change to make the property required if there is an x-prompt with no default,
and a little clean up to a) set up the basic option then b) check the xPrompt to set items and tooltips

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.html
```diff
@@ -2,7 +2,7 @@
   {{ field.description }}
```
sandikbarr, line 2: These form fields didn't have any awareness of the Validators set on the parent form group, so I connected them up with ControlContainer.

libs/vscode-ui/components/src/lib/select/select.component.spec.ts
```diff
@@ -1,21 +1,54 @@
+import { Component, ViewChild } from '@angular/core';
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { FormBuilder, ReactiveFormsModule } from '@angular/forms';
 import { OptionType } from '@angular/cli/models/interface';
 import { Option, OptionComponent } from '@nx-console/schema';
 import { SelectComponent } from './select.component';
```
sandikbarr, line 6: TODO: I am going to have to update the other vscode-ui-component specs and get them running in another pass on this.

libs/vscode-ui/components/src/lib/select/select.component.spec.ts
```diff
@@ -1,21 +1,54 @@
...
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ParentFormComponent],
+      imports: [ReactiveFormsModule]
+    }).compileComponents();
+  }));
+
   beforeEach(() => {
-    component = new SelectComponent();
```
sandikbarr, line 18: I have never understood why I sometimes see Angular component specs without TestBed. Totally fine I guess when it's just a presentational component with no dependencies.

libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.html
```diff
@@ -15,9 +15,11 @@
               [formControl]="filterFieldsControl"
             />
           </div>
-          <div class="run-button" (click)="runCommand(taskExecForm)">
```
sandikbarr, line 18: I don't see any reason why this needed to be `<div>` instead of button.

libs/vscode-ui/components/src/lib/multiple-select/multiple-select.component.html
```diff
@@ -1,20 +1,21 @@
...
   {{ field.description }}
 </div>
-
-<select
-  [formControl]="selectControl"
-  multiple
-  role="combobox"
-  [id]="field.name"
-  [required]="field.required"
-  [disabled]="disabled"
```
sandikbarr, line 11: Removed the disabled attribute and @Input() to clear the ReactiveForms warning about setting it on the FormControl instead. I can't see any reason why these need to support disabled right now.

apps/vscode/src/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "angular-console",
   "displayName": "Nx Console",
   "description": "Nx Console for Visual Studio Code",
-  "version": "16.0.2",
```
sandikbarr, line 5: Seems like a significant enough change to warrant a major version update.

libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.ts
```diff
@@ -331,8 +331,8 @@ export class TaskExecutionFormComponent implements OnInit, AfterViewChecked {
     dryRun?: boolean;
   }) {
     const flags = this.serializeArgs(form.value, architect);
+    flags.push('--no-interactive');
```
srlee309, line 334: Would it be better to use the defaults flag which disables interactive input prompts for options with a default. 

https://angular.io/cli/generate#options

sandikbarr: Hmm, yeah. Good question, nx should have this option too https://nx.dev/latest/angular/cli/generate#defaults

libs/vscode-ui/components/src/lib/select/select.component.spec.ts
srlee309: This is unrelated to the review I am just curious about your opinion on this. In my view,  across Angular projects there are a lot of component tests that gain no benefit from TestBed. TestBed also adds code to your test making it longer in terms of lines to read and time taken to run.. If you have a component spec file that isn't interacting with the template I prefer to use no TestBed.  When you do want to test the template, I have found storybook/cypress is a better way to do it than using TestBed. 

This is the related ngconf video on this topic: https://www.youtube.com/watch?v=7JucMlrs3dQ

sandikbarr: I can see your point. These tests wouldn't have benefitted from TestBed as they were. Storybook/Cypress really can be more beneficial for testing the template. 

...I broke the existing tests by adding the ControlContainer dependency and wanted to add a test target to get these running with the rest of them. Ran into some blockers though.

Cammisuli (approved):
Sandi and I went through this last week, and I tested it this morning. We'll merge this in and address the testing and other concerns with future improvements.

PR #1011, opened 2021-02-12
properly start debugging

Cammisuli:
This sets the tasks.json to "listen" to the terminal output so that it can start up the next task. This means that we can start debugging with just F5 and not have to do weird workarounds.


sandikbarr (approved):
tysm! 🎉

PR #1035, opened 2021-03-05
fix: unit tests on vscode-ui libs

sandikbarr:
get those specs running

### DIFF
--- a/libs/vscode-ui/components/src/lib/multiple-select/multiple-select.component.ts
+++ b/libs/vscode-ui/components/src/lib/multiple-select/multiple-select.component.ts
@@ -4,8 +4,6 @@ import {
   Input,
   Output,
   EventEmitter,
-  OnChanges,
-  SimpleChanges,
   OnInit,
   OnDestroy,
 } from '@angular/core';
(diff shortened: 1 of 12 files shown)


libs/vscode-ui/components/src/lib/multiple-select/multiple-select.component.ts
```diff
@@ -19,21 +17,23 @@ import { Subscription } from 'rxjs';
...
   parentFormGroup: FormGroup;
 
-  constructor(private readonly controlContainer: ControlContainer) {
-    this.parentFormGroup = this.controlContainer.control as FormGroup;
-  }
+  constructor(private readonly controlContainer: ControlContainer) {}
 
   private readonly subscriptioins = new Subscription();
 
   ngOnInit(): void {
```
sandikbarr, line 32: Found a little more cleanup on this form component... In the template, it was setting [formControl] and [formControlName] and formControl was being set to that empty initialized `selectControl = new FormControl([])`. That empty initialized FormControl seems to be why the ngOnChanges was doing the setValue below. No bueno. Now it grabs the form control from the parent form group by name and the value is set with the FormBuilder in the parent TaskExecutionFormComponent.

libs/vscode-ui/components/src/lib/select/select.component.html
```diff
@@ -4,10 +4,8 @@
 
 <div [formGroup]="parentFormGroup">
   <select
-    #selectControl
     (change)="updateValue($event.target.value)"
     role="combobox"
```
sandikbarr, line 8: With Reactive forms, you do not set [value], the control values are set on the FormGroup and connected with the formControl or formControlName directives.

libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.ts
```diff
@@ -249,11 +249,12 @@ export class TaskExecutionFormComponent implements OnInit, AfterViewChecked {
         );
         validators.push((control) => {
           if (
-            control.value &&
-            !validValueSet.has(control.value) &&
+            (control.value &&
+              !Array.isArray(control.value) &&
+              !validValueSet.has(control.value)) ||
             // multiselect values are Array, check if all values are in Set
-            control.value.length &&
```
sandikbarr, line 255: whoops, this was quite right. A valid control.value is either a string in the validValueSet or an array of values from validValueSet.

libs/vscode-ui/components/src/lib/multiple-select/multiple-select.component.ts
```diff
@@ -19,21 +17,23 @@ import { Subscription } from 'rxjs';
...
+  selectControl: FormControl;
   items: string[];
   parentFormGroup: FormGroup;
 
-  constructor(private readonly controlContainer: ControlContainer) {
-    this.parentFormGroup = this.controlContainer.control as FormGroup;
-  }
+  constructor(private readonly controlContainer: ControlContainer) {}
 
   private readonly subscriptioins = new Subscription();
```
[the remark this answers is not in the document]
sandikbarr: no problem. Honestly I debated just deleting this code because nothing actually uses the output this Subscription is set up for. All the components have change Output but none are using it because they were also all hooked into Reactive forms. I can see why you might want generic components that can do both Template and Reactive forms, but it's just never gonna be used here. Anyway, I'll fix the spelling for sure!

Cammisuli (approved):

PR #1036, opened 2021-03-05
fix: support directory as path alias from context menu

### LINKED ISSUE #997: VSCode Extension doesn't prefilled directory for @nwrl/nest:controller
When I right click on a directory in VSCode explorer, select Nx generate (UI) and then  @nwrl/nest controller, the "Directory" field stay empty. 

The controller is then generated in the src folder, not inside the inital clicked folder.

For exemple, on a right click on  .apps/api/src/app/domain, the controller will be, at src level.

sandikbarr:
Fixes #997 

It's not just the @nwrl/nest:controller generator. The directory property is also in @nrwl/node:application, @nrwl/node:library, and others.


Cammisuli (approved):
Thanks Sandi!

Cammisuli:
I'm not gonna merge this yet until we get the change log generator in. It's better to have the commit messages sorted out before we merge so that we don't have to rewrite commmits in the main branch 😅

PR #1038, opened 2021-03-10
perf: increase performance of nx console

Cammisuli:
This removes JSON5 parsing to use regular `JSON.parse`. There might be some json files that allow comments, so we have `stripJsonComments` included as well. 

This is the same behaviour that's in core Nx, so if there is any configuration file that has JSON5 specs in it, Nx will fail anyway ☹️ 

Here's a gif with these changes on a workspace that has 1426 projects with the following specs:
image

nx-console-perf-fix 2021-03-12 13_50_42


libs/server/src/lib/utils/utils.ts
```diff
@@ -131,7 +131,7 @@ export function fileExistsSync(filePath: string): boolean {
 }
 
 export function readAndParseJson(fullFilePath: string): any {
-  return JSON5.parse(readFileSync(fullFilePath).toString());
+  return JSON.parse(stripJsonComments(readFileSync(fullFilePath).toString()));
```
sandikbarr, line 134: mentioned this to Isaac and he wondered if it should also try catch and fallback to JSON5 if the regular parse fails

Cammisuli: This is the same "technique" used in Nx. If the workspace.json file includes anything related to json5 (trailing commas, etc) then Nx itself won't work anyway ☹️ 

I still want to work on this though, I think making this asynchronous will help with other things.

isaacplmann:
It could be that this is present to handle plain angular CLI configs as well.

apps/vscode/src/package.json
```diff
@@ -37,7 +37,7 @@
     "Other"
   ],
   "activationEvents": [
-    "*"
+    "workspaceContains:nx.json"
```
Cammisuli, line 40: I also made this only activate when you have a nx.json file in your vscode workspace. This should free up memory for folks who use vscode for things that aren't Nx 👀

sandikbarr (approved):
🎉

PR #1039, opened 2021-03-12
refactor: create lib projects

Cammisuli:
## What it does

This PR is essentially going to create node libs for each part fo the main `vscode` app. This will better reflect what we expect out of Nx workspaces.  

### Investigate

I'm not sure if it's the product of this refactor, but things appear to be loading slower... I might need to continue with https://github.com/nrwl/nx-console/pull/1041 after.


sandikbarr (approved):
HUGE 👍 👍

PR #1040, opened 2021-03-12
chore: create release tools to generate change log and bump verisons

Cammisuli:
Uses the `standard-version` tool to read commits and automatically bump versions and create git tags. 

This is done locally rather than a CI. Follow up PR to deploy the extension whenever a new version tag is pushed to the repo


sandikbarr (approved):
💯

Issue #965, opened 2020-07-29
Remember location of workspace.json/angular.json in subfolder


Darkmagister:
Hello, i don't know if it's an issue but i have a question about nx console vscode module, i have a nodejs and Angular project (they are connected), so i have a project directory with inside 2 Dir, one for angular (project/angular) and one for node (project/node) i usually open the project directory in vsCode but every time i open it and i go to the nx console plugin to serve it it ask me to select the workspace, and i have to navigate to the project/angular directory and then it work great, is there any configuration or file i have to create to make nx console find the correct workspace without having to tell it every time ?

thanks a lot

davidzwa:
I have the same feature request.

The NX Console plugin tries to find either `nx.json` `workspace.json` or `angular.json` in the root folder if I'm correct. That's what missing here. They expect us to open the project folder directly, and not a subfolder.

I work in a .NET Core solution, so a subfolder is almost always default. I would like to see a VS Code `settings.json` setting to patch this.

daton89:
I'm also having this issue, I'm working with angular and express, in a repo with client and server, and I migrated from angularCli actually with the client folder, and I need to select the workspace each time I open vscode, would be really great to have the possibility to store the workspace path in the settings.

Cammisuli:
With the recent changes I did with #1039, I changed the way we store the workspace file path. This means that it should be stored in a VSCode workspace context, and we should be able to read from there on extension activation.

I'll have to play around with it, but the foundation should be there shortly 😃

PR #1052, opened 2021-03-26
feat: load custom workspace path on extension load

Cammisuli:
## What it does
Loads the custom workspace path when the extension is loaded. This path is stored per vscode workspace. 

I also show a description of what the current workspace is:
image

Fixes: #965


sandikbarr (approved):
super

PR #1053, opened 2021-03-26
feat: only execute dry run if Generate form is valid

sandikbarr:
small change to only do the dryRun runCommand for Generate if the form is valid


Cammisuli (approved):
Nice! 😄

PR #1056, opened 2021-03-31
chore: better debugging for webview (ng app)

Cammisuli:
## What it does
This change allows for better debugging with the webview in the extension. Summary of changes:
* inline source maps for better breakpoints
* when the main.js file changes, the webview html will be re-rendererd


sandikbarr (approved):
life changing 14/10

PR #1058, opened 2021-04-01
fix: refresh projects view clears cache

### LINKED ISSUE #1042: Changing workspace/angular.json does not update projects
When adding new builders or projects to the workspace, the projects view does not update to show the new targets or projects. 

The json cache needs to be busted so that refreshes can get the latest projects and options. 

> Most likely need to include a auto refresh to watch workspace.json files.

Cammisuli:
## What it does
Clicking the refresh icon in the Projects view will delete the cache for the current workspace file. 

This PR also adds a mechanism to watch the workspace file and automatically refresh the projects view. 

Closes #1042

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,6 +11,7 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "args": [
+        "--disable-extensions",
         "--extensionDevelopmentPath=${workspaceFolder}/dist/apps/vscode"
       ],
       "trace": "false",
--- a/apps/vscode/src/package.json
+++ b/apps/vscode/src/package.json
@@ -55,11 +55,6 @@
       "view/title": [
         {
           "command": "nxConsole.refreshNxProjectsTree",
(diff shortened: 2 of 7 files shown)


.vscode/launch.json
```diff
@@ -11,6 +11,7 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "args": [
+        "--disable-extensions",
```
Cammisuli, line 14: I added this here so that we just get errors related to our extension 😅

apps/vscode/src/package.json
```diff
@@ -82,6 +77,10 @@
         }
       ],
       "commandPalette": [
+        {
+          "command": "nx.generate.ui.fileexplorer",
+          "when": "false"
+        },
```
Cammisuli, line 83: Not related to this PR, but I hid this command from the command palette as it doesn't make sense there.

sandikbarr (approved):
This is working great for me. Thanks for fixing this.

PR #1059, opened 2021-04-07
feat: shows argument list and highlights changed and invalid fields in menu

sandikbarr:
Re: #1022
* Command arguments are shown in the right side gutter (which also reduces width on property descriptions)
* Adds `Copy to clipboard` button for nx/ng task command
* Reduce horizontal margin on fields for higher content density
* Left side fields nav menu 
  * properties that have been modified are highlighted in the menu
  * properties that are invalid are highlighted as errors in the menu
Screen Shot 2021-04-29 at 9 37 18 AM
Screen Shot 2021-04-29 at 9 37 49 AM


isaacplmann:
I like this idea!  Probably need to double check the colors with a light theme too.

sandikbarr:
> I like this idea! Probably need to double check the colors with a light theme too.

Good call. We use the theme colors `--vscode-errorForeground` for error text and I found the `--vscode-gitDecoration-modifiedResourceForeground` theme color for modified because I didn't see a foreground color in the regular `--vscode` theme colors that really fit. So that modified color would match the color of any git tracked changed files.

In this light theme screenshot, the Name field has been modified and is valid, and the Implements and Project fields have been touched and have errors. The Project field is also bold because there is existing support that highlights the scroll position in the menu.
Screen Shot 2021-04-08 at 8 47 11 AM

sandikbarr:
Feedback from @jaysoo is that we should make sure you can copy and paste the command with the arguments. With the layout as is with the command in the header and the arguments in the right sidebar, they can't be selected and copy-pasted in one go. I didn't want to put the arguments in the header because that would push down the form as you're editing, so we are thinking a copy button that would copy the command along with the arguments would be a good solution. I'll work on that before this feature gets merged.

libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.spec.ts
```diff
@@ -5,6 +5,8 @@ import { waitForAsync, ComponentFixture, TestBed } from '@angular/core/testing';
 import { TaskExecutionSchema } from '@nx-console/schema';
 import { TASK_EXECUTION_SCHEMA } from './task-execution-form.schema';
 import { VscodeUiComponentsModule } from '@nx-console/vscode-ui/components';
```
sandikbarr, line 7: I can't figure out why VS Code shows this lint error on these [ArgumentListModule imports](https://github.com/nrwl/nx-console/pull/1059/files#diff-37a9daab57c623d442d8f82f28baadfdd1b1db33e7c77838aa99c82b2c92237aR8). It does't fail `nx lint` and another weird things is VS Code allows it if I change it to a relative import (which doesn't pass lint obvs but wtf).

Also the vscode-ui project is in apps and nothing is being imported from it here. Kinda feels like the partial matching on this rule isn't right.
```
Circular dependency between "vscode-ui-feature-task-execution-form" and "vscode-ui" detected: vscode-ui-feature-task-execution-form -> vscode-ui -> vscode-ui-feature-task-execution-formeslint@nrwl/nx/enforce-module-boundaries

File: libs/vscode/webview/src/lib/webview.ts [hunk 2]
```diff
@@ -178,6 +185,7 @@ export function getIframeHtml(
       window.vscode = acquireVsCodeApi();
     </script>
     <link href="${webView.asWebviewUri(stylePath)}" rel="stylesheet"/>
+    <link href="${webView.asWebviewUri(codiconsPath)}" rel="stylesheet" />
```
sandikbarr, line 188: I'm occasionally seeing a 404 on this codicon.css in the console, might just be showing up on dev mode rebuild but looking into it.

libs/vscode-ui/components/src/lib/vscode-ui-components.module.ts
```diff
@@ -8,7 +8,6 @@ import { CheckboxComponent } from './checkbox/checkbox.component';
 import { FieldComponent } from './field/field.component';
 import { FieldTreeComponent } from './field-tree/field-tree.component';
 import { MultipleSelectComponent } from './multiple-select/multiple-select.component';
-import { FormatTaskPipe } from './format-task/format-task.pipe';
```
sandikbarr, line 11: This pipe is used over in the parent form, and it needs to call the formatTask function directly now as well for the copy to clipboard functionality, so I moved it into that lib.

libs/vscode/webview/src/lib/webview.ts
sandikbarr: Resolved by adding vscode-codeicons to vscode app package.json dependencies and referencing in the dist node_modules

libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.spec.ts
sandikbarr: Whatever was causing this went away, so cool 😎

libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.scss
```diff
@@ -28,7 +28,6 @@ $offset-with-configurations: $offset-without-configurations +
   border: $input-border-width solid var(--vscode-settings-textInputBorder);
   color: #fff;
   text-align: center;
-  line-height: 30px;
```
sandikbarr, line 31: fixes this jankiness on the Run button focus border 
Screen Shot 2021-04-29 at 9 42 35 AM
fixed:
Screen Shot 2021-04-29 at 9 43 30 AM

Cammisuli: I hate when that happens 😅

Cammisuli (approved):
Nice! Thanks Sandi!

PR #1060, opened 2021-04-09
feat: vscode-test setup

sandikbarr:
...so far just gets vscode-test running using the example test


sandikbarr:
eh, we got a little further on this but kinda wishing there were [more to it](https://code.visualstudio.com/api/working-with-extensions/testing-extension)

PR #1062, opened 2021-04-15
feat: updates form components to support long form items plus specs 

### LINKED ISSUE #984: Autocomplete component doesn't retain the filtered options in the list when you select an option or click outside after typing in the input.
The autocomplete doesn't retain the filtered options in the list when you select an option or click outside after typing in the input. It should continue to filter the options by the item in the input after clicking outside or selecting an option. 

autocomplete filtered options not retained

sandikbarr:
* The `vscode-ui-feature-task-execution-form-e2e` was failing because the tests expected support for long form options, which was not consistently supported across the dropdown components. Should be fixed.
* Fixes #984 to retain autocomplete filtered options when it loses focus
* Many more specs 🎉  Feels nice to have some narrower specs next to the components.

### DIFF
--- a/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
+++ b/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
@@ -95,7 +95,7 @@ export class AutocompleteComponent implements OnInit, ControlValueAccessor {
     this.visibleOptions = this._options$.pipe(
       switchMap((options) =>
         this.control.valueChanges.pipe(
-          startWith(''), // When panel is first opened, show the entire list
+          startWith(this.control.value || ''), // When panel is first opened, show the entire list
           map(
             (formValue) =>
               options &&
--- a/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
+++ b/libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
@@ -1,9 +1,6 @@
 import {
   Component,
-  OnDestroy,
   Input,
-  SimpleChanges,
-  OnChanges,
   ViewChild,
   ElementRef,
   OnInit,
@@ -19,7 +16,6 @@ import {
 import {
   map,
   tap,
-  takeUntil,
   switchMap,
   startWith,
   scan,
@@ -27,15 +23,9 @@ import {
   filter,
   shareReplay,
 } from 'rxjs/operators';
-import {
-  Subject,
-  BehaviorSubject,
-  Observable,
-  fromEvent,
-  merge,
-  of,
-} from 'rxjs';
-import { ItemsWithEnum, Option } from '@nx-console/schema';
+import { BehaviorSubject, Observable, fromEvent, merge, of } from 'rxjs';
+import { Option } from '@nx-console/schema';
+import { getOptionItems } from '../field-items/field-items.pipe';
 
 export enum AutocompleteNavKeys {
   Enter = 'Enter',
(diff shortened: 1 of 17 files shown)


libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -58,52 +48,40 @@ export const AUTOCOMPLETE_NAV_KEYS = new Set(['Enter', 'ArrowUp', 'ArrowDown']);
     },
   ],
 })
-export class AutocompleteComponent
-  implements OnInit, OnDestroy, OnChanges, ControlValueAccessor {
-  @Input() field: Option;
-
+export class AutocompleteComponent implements OnInit, ControlValueAccessor {
```
sandikbarr, line 51: I don't really think this needs to be a ControlValueAccessor with the input connected thru ControlContainer, but I'll save that for later. Same with FieldComponent; seems like it should just be a pass thru to child form components using the built in ControlValueAccessors.

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -114,6 +92,21 @@ export class AutocompleteComponent
       )
     ).pipe(startWith(false), debounceTime(300), shareReplay(1));
 
+    this.visibleOptions = this._options$.pipe(
+      switchMap((options) =>
+        this.control.valueChanges.pipe(
```
sandikbarr, line 97: * moved all setup to OnInit
* Fixed #984  with this change from `startWith('')` to `startWith(this.control.value || '')`

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -114,6 +92,21 @@ export class AutocompleteComponent
...
+              options &&
+              options.filter((option) =>
+                option.toLowerCase().includes(formValue.toLowerCase())
+              )
+          ),
+        )
+      )
+    );
+
     this.selectedOption$ = this.isFocused$.pipe(
```
sandikbarr, line 110: `selectedOption$` is a little bit 😬 but I left it alone. I did remove the subscribe to it below though.

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -152,36 +145,21 @@ export class AutocompleteComponent
       ),
       shareReplay(1)
     );
-
-    this.selectedOption$.subscribe();
-  }
-
-  ngOnChanges(changes: SimpleChanges) {
-    if (changes.field) {
```
sandikbarr, line 160: move this to `set field()`

libs/vscode-ui/components/src/lib/field-items/field-items.pipe.ts
```diff
@@ -0,0 +1,19 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { ItemsWithEnum, Option } from '@nx-console/schema';
+
+export const getOptionItems = (field: Option): string[] | undefined => {
+  return (
```
sandikbarr, line 5: this typing can be a little annoying on items, especially to access these items in the template, so the pipe

libs/vscode-ui/components/src/lib/field/field.component.ts
```diff
@@ -46,9 +47,6 @@ export class FieldComponent implements ControlValueAccessor, OnDestroy {
   valueChangeSub: Subscription;
   OptionComponent = OptionComponent;
 
-  control = new FormControl('');
```
sandikbarr, line 49: finally removing all of this new FormControl down in these components and just connecting thru the ControlContainer 😸

libs/vscode-ui/components/src/lib/multiple-select/multiple-select.component.ts
```diff
@@ -21,46 +21,25 @@ export class MultipleSelectComponent implements OnInit, OnDestroy {
...
-    this.parentFormGroup = this.controlContainer.control as FormGroup;
-    this.selectControl = this.parentFormGroup.get(
-      this.field.name
-    ) as FormControl;
     this.subscriptions.add(
       this.selectControl.valueChanges.subscribe((value) =>
         this.valueChange.emit(value)
       )
     );
-    if (this.field.items) {
```
sandikbarr, line 42: Another significant change I made is to not set items OnInit but instead just access them from the field.
OnInit was fine given the form is created each time a new generator or executor is opened, but it was a pain to build tests around. Now the fields update OnChange as you would expect.

libs/vscode-ui/components/src/lib/select/select.component.spec.ts
```diff
@@ -48,23 +58,47 @@ describe('SelectComponent', () => {
...
+    it('should ignore when there is no tooltip for option value', () => {
+      const noTooltipOption = options.find(
+        (opt) => opt.nativeElement.textContent.trim() === 'other'
+      );
+      expect(noTooltipOption).toBeDefined();
+      expect(noTooltipOption?.nativeElement.getAttribute('title')).toBeNull();
     });
   });
+
+  it('should show long form options', () => {
```
sandikbarr, line 88: the SelectComponent wasn't supporting long form items

libs/vscode-ui/components/src/lib/autocomplete/autocomplete.component.ts
Cammisuli: Yea, that subscribe is a weird pattern..

libs/vscode-ui/components/src/lib/field-items/field-items.pipe.ts
Cammisuli: Nice!

Cammisuli (approved):
Everything looks good for me! I tested it locally, and things are good :D

PR #1064, opened 2021-04-19
feat: run-many prompts for projects

sandikbarr:
* On the Nx: run-many command, when selecting projects it now prompts for selection from a quick pick list of projects with the given target instead of having the user type out the project entries.
* The target flag has also been included in the userSetFlags in order to show the target with the command being built up.
run-many projects quickpick from target

### DIFF
--- a/libs/vscode/tasks/src/lib/nx-task-commands.ts
+++ b/libs/vscode/tasks/src/lib/nx-task-commands.ts
@@ -142,11 +142,6 @@ const AFFECTED_OPTIONS: Option[] = [
 ].map((v) => ({ ...v, aliases: [] }));
 
 const RUN_MANY_OPTIONS: Option[] = [
-  {
-    name: 'projects',
-    type: OptionType.Array,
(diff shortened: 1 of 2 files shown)


libs/vscode/tasks/src/lib/nx-task-commands.ts
```diff
@@ -266,14 +261,28 @@ async function promptForRunMany() {
...
+        type: OptionType.Array,
+        description: 'Projects to run',
+        aliases: [],
+        enum: projects,
+      },
+      ...RUN_MANY_OPTIONS,
+    ];
+  }
+
+  const flags = await selectFlags('run-many', options, 'nx', { target });
```
sandikbarr, line 279: Adding target to the userSetFlags parameter shows 'Execute: nx run-many --target=<target>` on the Command Palette instead of just `nx: run-many` with the selected options. Seems to make more sense as that's how you'd use the run-many command.

Cammisuli (approved):

PR #1065, opened 2021-04-20
feat: nx run in command palette and context menu

sandikbarr:
Adds support for `nx run` in the Command Palette
Prompts for a project then a target and supported flags (including configuration with a dropdown of the configuration options.)
run command

### DIFF
--- a/libs/vscode/tasks/src/lib/cli-task-commands.ts
+++ b/libs/vscode/tasks/src/lib/cli-task-commands.ts
@@ -49,6 +49,10 @@ export function registerCliTaskCommands(
     );
   });
 
+  commands.registerCommand('nx.run', () =>
+    selectCliCommandAndPromptForFlags('run')
+  );
+
   commands.registerCommand(`ng.generate`, () =>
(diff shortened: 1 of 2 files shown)


libs/vscode/tasks/src/lib/cli-task-commands.ts
```diff
@@ -49,6 +49,10 @@ export function registerCliTaskCommands(
     );
   });
```
sandikbarr, line 51: I am not really sure of the distinction between nx-task-commands and cli-task-commands. It feels like this cli-task-commands was originally angular CLI support vs Nx CLI. I started out putting the run command support in nx-task-commands, but the flow fit in a little better here with the other CLI commands and it has all the support for getting the executor/builder definition and prompting for flags.

Cammisuli: yea, I noticed that as well. We'll eventually consolidate the two and just use Nx.

Cammisuli (approved):

PR #1066, opened 2021-04-22
fix: enable generate from context menu setting take effect immediately

sandikbarr:
With how the ["when" clause](https://code.visualstudio.com/api/references/when-clause-contexts) context for `isGenerateFromContextMenuEnabled` is just getting set on activate, this setting isn't applying immediately like other VS Code settings do. Seems like you have to reload the window to get the setting to work. 

But the "when" clause can read from the `config` settings in order to toggle it on/off in your current VS Code session, so this change should fix that.

### DIFF
--- a/apps/vscode/src/package.json
+++ b/apps/vscode/src/package.json
@@ -47,7 +47,7 @@
     "menus": {
       "explorer/context": [
         {
-          "when": "isGenerateFromContextMenuEnabled",
+          "when": "isAngularWorkspace && config.nxConsole.enableGenerateFromContextMenu || isNxWorkspace && config.nxConsole.enableGenerateFromContextMenu",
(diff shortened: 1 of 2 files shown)


apps/vscode/src/package.json
```diff
@@ -47,7 +47,7 @@
     "menus": {
       "explorer/context": [
         {
-          "when": "isGenerateFromContextMenuEnabled",
+          "when": "isAngularWorkspace && config.nxConsole.enableGenerateFromContextMenu || isNxWorkspace && config.nxConsole.enableGenerateFromContextMenu",
```
sandikbarr, line 50: doesn't support parens with the || and &&, so the condition is a bit long

Cammisuli (approved):

PR #1067, opened 2021-05-03
feat: Populates form with workspace defaults for build, e2e, lint, serve, and test targets

### LINKED ISSUE #930: fileReplacements option is missing in the production build
When I execute a production build in an Angular project, Angular Console states that it will run `ng build --prod` but internally it executes the command, passing the configuration options of the production one by one. Unfortunately, this does not take into account `fileReplacements` array so I end up with a _not so real_ production version of my app because environment files will not be correctly replaced.

sandikbarr:
* Reads executor options from the project in workspace
* Populates web form with workspace defaults for build, e2e, lint, serve, and test targets
* Fixes #930: Runs command with configuration flag instead of adding each configuration option as a flag

Screen Shot 2021-05-03 at 12 22 58 PM


Cammisuli:
woo 🎉  Thanks Sandi! 😄 

I'll review it as soon as I can.

Cammisuli (approved):
Awesome!

PR #1068, opened 2021-05-06
feat: load workspace defaults into Run executor form

sandikbarr:
* Loads the workspace defaults into the Run executor form
* Supports the configuration tab in the Run form when targets have one or more configurations
Screen Shot 2021-05-06 at 1 54 56 PM

### DIFF
--- a/libs/vscode/webview/src/lib/get-task-execution-schema.ts
+++ b/libs/vscode/webview/src/lib/get-task-execution-schema.ts
@@ -3,7 +3,6 @@ import {
   getOutputChannel,
   getTelemetry,
   readArchitectDef,
-  readBuilderSchema,
   selectSchematic,
 } from '@nx-console/server';
 import {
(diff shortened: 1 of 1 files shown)


libs/vscode/webview/src/lib/get-task-execution-schema.ts
```diff
@@ -94,24 +93,24 @@ export async function getTaskExecutionSchema(
             return;
           }
 
-          const builderOptions = await readBuilderSchema(
-            cliTaskProvider.getWorkspacePath(),
-            selection.architectDef.builder
```
sandikbarr, line 99: Gets the defaults: Stepping out to `verifyBuilderDefinition` similar to the other executors here where it reads the workspace and is including the target config options in the call to readBuilderSchema.

libs/vscode/webview/src/lib/get-task-execution-schema.ts
```diff
@@ -94,24 +93,24 @@ export async function getTaskExecutionSchema(
...
+            json
           );
-
-          if (!builderOptions) {
+          if (!validBuilder) {
             return;
           }
 
           return {
             ...readArchitectDef(
```
sandikbarr, line 106: Gets configurations: Arguments weren't quite right here. It was passing for example `readArchitectDef('run', 'build', 'store')` but it needs the architectDef from the selected target to get the configurations.

Cammisuli (approved):
Thank you Sandi!

PR #1069, opened 2021-05-11
feat: welcome view and getting started walkthrough

sandikbarr:
A recent VS Code update added experimental support for [Welcome Page walkthroughs](https://code.visualstudio.com/updates/v1_56#_welcome-page-walkthroughs).

I am only able to get the walkthrough to launch on installation in Insiders with the setting `"workbench.welcomePage.experimental.extensionContributions": true`.

But without Insiders, we still get this new button on the Welcome View to launch our walkthough:
Screen Shot 2021-07-30 at 3 50 01 PM

The walkthrough docs were copied from nx.dev, minus the videos. I could not get external links to work from within the walkthrough so wasn't able to include those.
Screen Shot 2021-07-30 at 3 50 33 PM

(just some previous research notes:)
> I also checked out what [GitLens](https://github.com/eamodio/vscode-gitlens) is doing with their welcome and latest updates content, which is a different direction we may want to explore for a more robust solution. They store the latest installed version, and when someone has installed it for the first time they show a Welcome Page custom webview, and when someone has just updated to a new version, they show a Latest Updates custom webview.

### DIFF
--- a/apps/vscode/src/package.json
+++ b/apps/vscode/src/package.json
@@ -621,6 +621,16 @@
         }
       ]
     },
+    "viewsWelcome": [
+      {
+        "view": "nxConsole.views.welcome",
+        "contents": "Nx Console - The UI for Nx"
+      },
+      {
+        "view": "nxConsole.views.welcome",
+        "contents": "[Getting Started](https://nx.dev/latest/angular/getting-started/console#nx-console-for-vs-code)"
+      }
+    ],
     "views": {
       "nx-console": [
         {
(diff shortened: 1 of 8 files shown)


apps/vscode/src/package.json
```diff
@@ -587,7 +647,15 @@
           "id": "nxProjects",
           "name": "Projects",
           "when": "isNxWorkspace || isAngularWorkspace"
-        }
```
sandikbarr, line 590: meh, got a ["welcome view"](https://code.visualstudio.com/api/references/contribution-points#contributes.viewsWelcome) showing up along with the nearly empty 'Select workspace' treeview when an angular or Nx project isn't open. Might work for "what is this?"
Screen Shot 2021-05-11 at 2 49 06 PM

Cammisuli: Can we move this up above the generate and run target view?

Cammisuli (changes requested):

apps/vscode/src/package.json
sandikbarr: Yeah, looks like we can! Great idea

Cammisuli (approved):

Issue #925, opened 2020-01-22
Extension causes high cpu load


vikicoder007:
- Issue Type: `Performance`
- Extension Name: `angular-console`
- Extension Version: `10.0.0`
- OS Version: `Windows_NT x64 10.0.18362`
- VSCode version: `1.41.1`

:warning: Make sure to **attach** this file from your *home*-directory:
:warning:`C:\Users\vikic\nrwl.angular-console-unresponsive.cpuprofile.txt`

Find more details here: https://github.com/microsoft/vscode/wiki/Explain-extension-causes-high-cpu-load

RobertYost:
I am also experiencing this, and unfortunately had to disable the extension - it added nearly 3 seconds to startup time alone and made all of vscode incredibly laggy.

Crocetti:
I am also experiencing this.
[nrwl.angular-console-unresponsive.cpuprofile.txt](https://github.com/nrwl/nx-console/files/4492548/nrwl.angular-console-unresponsive.cpuprofile.txt)

Abdallah-khalil:
[nrwl.angular-console-unresponsive.cpuprofile.txt](https://github.com/nrwl/nx-console/files/4808552/nrwl.angular-console-unresponsive.cpuprofile.txt)
Having the same when trying to open the generate

Cammisuli:
Thank you for the report! I've definitely noticed a lot of performance issues and I want to fix those up. I'll keep these in mind when I'm going through a performance rundown.

PR #1074, opened 2021-05-17
fix: only reads and normalizes selected schematic options

sandikbarr:
Improves performance for `Generate` form and `nx generate` within command palette by only reading and normalizing options for the selected schematic.

Fixes: #925


Cammisuli:
This will fix this issue #925

Cammisuli (approved):

PR #1075, opened 2021-05-21
feat: workspace codelens

sandikbarr:
adds nx run targets CodeLens to the workspace config


apps/vscode/src/main.ts
```diff
@@ -42,6 +45,10 @@ import {
   NxProjectTreeItem,
   NxProjectTreeProvider,
 } from '@nx-console/vscode/nx-project-view';
```
sandikbarr, line 47: I moved `verifyWorkspace` from from the `verify` lib to the `nx-workspace` lib to resolve a circular dependency between the two.

libs/vscode/nx-workspace/src/lib/workspace-codelens-provider.ts
```diff
@@ -0,0 +1,89 @@
...
+    super(range);
+  }
+}
+export class WorkspaceCodeLensProvider implements CodeLensProvider {
+
+  /**
+   * Provides a CodeLens set for a matched document
+   * @param document a document matched by the pattern passed to registerCodeLensProvider
+   * @returns ProjectCodeLens Range locations and properties for the document
+   */
```
sandikbarr, line 23: took out async on this function because it was no longer returning a Promise

libs/vscode/tasks/src/lib/cli-task-commands.ts
```diff
@@ -49,8 +47,18 @@ export function registerCliTaskCommands(
     );
   });
 
-  commands.registerCommand('nx.run', () =>
-    selectCliCommandAndPromptForFlags('run')
+  commands.registerCommand(
+    'nx.run',
+    (project?: string, target?: string, configuration?: string) => {
+      let flags;
+      if (configuration) {
```
sandikbarr, line 54: I went with `-c configuration` here because `ng run` on an angular cli project doesn't take --prod as an argument, but it does accept -c. Technically, the CodeLens is formatted as `nx|ng run project:target:configuration` which is likely the most preferred form, but I wanted to pass configuration as flags and run it in the same fashion as these other cli task commands.

apps/vscode/src/main.ts
```diff
@@ -259,3 +269,37 @@ function registerWorkspaceFileWatcher(
...
+        );
+        if (enableWorkspaceConfigCodeLens && !codeLensProvider) {
+          registerWorkspaceCodeLensProvider(context);
+        } else if (!enableWorkspaceConfigCodeLens && codeLensProvider) {
+          codeLensProvider.dispose();
+          codeLensProvider = null;
+        }
+      }
+    })
+  );
```
Cammisuli, line 304: Nice! This looks like it works properly :smile:

Do you think this logic should be handled in the workspace lib? Like have a main entry to that lib, and then do all this logic in there?

libs/vscode/nx-workspace/src/lib/workspace-codelens-provider.ts
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Resolves and sets the command on visible CodeLens
+   * @param lens lens to be resolve
+   * @returns ProjectCodeLens with command
+   */
+  // https://github.com/microsoft/vscode-extension-samples/blob/main/codelens-sample/src/CodelensProvider.ts
+  resolveCodeLens(lens: CodeLens): CodeLens | Promise<CodeLens> | null {
+    if (lens instanceof ProjectCodeLens) {
+      const command: Command = {
+        command: 'nx.run',
```
Cammisuli, line 78: Does this need to change to `'ng.run'` for angular workspaces?

sandikbarr: I didn't actually implement ng.run 😬 because I didn't realize at the time we'd also be able to use that with angular cli workspaces as well. I should add it and use it here as well though; won't be too much. Good point.

apps/vscode/src/main.ts
sandikbarr: That makes sense. Can't keep throwing everything in main/activate.

libs/vscode/tasks/src/lib/cli-task-commands.ts
```diff
@@ -49,30 +47,30 @@ export function registerCliTaskCommands(
...
-    selectCliCommandAndShowUi('generate', context.extensionPath)
-  );
-  commands.registerCommand(`nx.generate`, () =>
-    selectSchematicAndPromptForFlags()
-  );
+    commands.registerCommand(`${cli}.generate.ui`, () =>
+      selectCliCommandAndShowUi('generate', context.extensionPath)
+    );
+
+    commands.registerCommand(`${cli}.generate.ui.fileexplorer`, (uri: Uri) =>
```
sandikbarr, line 69: The context menu generate and run commands were also just implemented in the `Nx` category. All of our other commands are implemented both as `nx` and `ng`.
* Right click `Nx generate (ui)` was working on Angular CLI projects, but it was just an "nx" VS Code command. 
* Adding a separate command for `ng.generate.ui.fileexplorer` gives us finer control over that menu item, so I made it "ng generate (ui)" for Angular CLI projects.
* I also added `ng run` to the context menu for Angular CLI projects that uses the Command Palette and prompts

libs/vscode/tasks/src/lib/cli-task-commands.ts
```diff
@@ -105,7 +103,19 @@ function selectCliCommandAndShowUi(
...
 
-async function selectCliCommandAndPromptForFlags(command: string, projectName?: string) {
+async function selectCliCommandAndPromptForFlags(
+  command: string,
+  projectName?: string,
+  target?: string,
+  configuration?: string
+) {
+  let flags: string[] | undefined;
+  if (configuration) {
```
sandikbarr, line 113: Ok, I actually finally settled on `--configuration=production` because that will resolve the same as the other `run` command tasks to give us that "Task is already active" if something is already running with the same command and args.

apps/vscode/src/main.ts
```diff
@@ -117,6 +124,9 @@ export function activate(c: ExtensionContext) {
       manuallySelectWorkspaceDefinitionCommand
     );
 
+    // registers itself as a CodeLensProvider and watches config to dispose/re-register
```
sandikbarr, line 127: How's this? I have this registering itself as a CodeLensProvider and watching the config settings for changes to dispose/re-register?

sandikbarr:
Thank you for your feedback on this @Cammisuli. I've been using it on a few different workspaces and am feeling pretty confident in it, but lmk if you see anything else that should be addressed. 🎉

PR #1077, opened 2021-05-31
feat: provide json schema for workspace.json

Cammisuli:
## What it does
Adds a json schema file to provide intellisense when configuring the workspace.json (or angular.json) file.

It will read the package.json file and find the packages that contain builder or executor configurations. The json schema file will then map each executor to it's options. 

Here's how it looks with the `angular.json` file in this project:
image
image

## Notes

Right now I only handle the targets/architects. We could add support to provide support for the schematics/generators section as well.

### TODO
Add support for configuration sections


sandikbarr (approved):
This is SO freakin cool! 💯 🚀

PR #1082, opened 2021-06-11
fix: strip appsDir and libsDir from path if app|application or lib|library

sandikbarr:
I can't seem to figure out how we could determine if a schematic/generator is a project and what the projectType is from the schema before it is generated, and I am not totally happy with this `schematic.name === 'application' || schematic.name === 'app'` and `schematic.name === 'library' || schematic.name === 'lib'` solution. 

Update: added a commit that only sets project/projectName or path/directory from the explorer context menu, not both. Project takes precedence.

### DIFF
--- a/libs/vscode/webview/src/lib/get-task-execution-schema.ts
+++ b/libs/vscode/webview/src/lib/get-task-execution-schema.ts
@@ -1,4 +1,4 @@
-import { TaskExecutionSchema } from '@nx-console/schema';
+import { Option, TaskExecutionSchema } from '@nx-console/schema';
 import {
   getOutputChannel,
   getTelemetry,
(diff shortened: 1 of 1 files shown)


libs/vscode/webview/src/lib/get-task-execution-schema.ts
```diff
@@ -170,21 +172,37 @@ export async function getTaskExecutionSchema(
...
-    const path = contextMenuUri.fsPath
-      .replace(cliTaskProvider.getWorkspacePath(), '')
+
+    const workspacePath = cliTaskProvider.getWorkspacePath();
+    let path = contextMenuUri.fsPath
+      .replace(workspacePath, '')
       .replace(/\\/g, '/')
       .replace(/^\//, '');
 
+    const nxConfig = readAndCacheJsonFile('nx.json', workspacePath);
```
sandikbarr, line 189: I tried this with `workspaceLayout()` using a deep import from nx/workspace, but it wasn't giving me the configured path with a workspace I was testing on for some reason.

libs/vscode/webview/src/lib/get-task-execution-schema.ts
```diff
@@ -170,21 +172,37 @@ export async function getTaskExecutionSchema(
...
+    if (libsDir && schematic.name === 'library' || schematic.name === 'lib') {
+      path = path.replace(libsDir, '')
+        .replace(/^\//, '');
+    }
+
     return {
       path,
       directory: path,
       project: projectName,
+      projectName,
```
sandikbarr, line 205: I also slipped in support for setting the `projectName` schema option. Some of our generators use `projectName` rather than `project`.

Cammisuli (approved):
Nice, things are working better now!

PR #1084, opened 2021-06-16
ci(github): updated actions

Phillip9587:
- Updated github actions to their latest versions
- Cache the Yarn Cache Directory instead of node_modules
- Removed Matrix from Publish workflow

### DIFF
--- a/.github/workflows/ci_checks.yml
+++ b/.github/workflows/ci_checks.yml
@@ -2,30 +2,37 @@ name: CI Checks
 on: [push, pull_request]
 
 jobs:
-  Checks:
+  checks:
     runs-on: ${{ matrix.os }}
+
     env:
       NX_BRANCH: ${{ github.event.number || github.ref }}
       NX_RUN_GROUP: ${{ github.run_id }}
+
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [14.x]
+        node-version: [14]
+
     steps:
+      - uses: actions/checkout@v2
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
-      - uses: actions/checkout@v2
-      - uses: actions/cache@v1
-        id: workspace-cache
+
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_checks.yml
```diff
@@ -2,33 +2,40 @@ name: CI Checks
...
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - uses: actions/cache@v2
         with:
-          path: node_modules
-          key: ${{ runner.os }}-${{ matrix.node-version }}-workspace-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-${{ matrix.node-version }}-workspace-
+          path: ${{ steps.yarn-cache-dir.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-yarn-
```
Cammisuli, line 31: I liked caching the whole node_modules directory because doing `yarn install` resolves quicker. 
It makes me feel like things are working properly because I see that the yarn install step happens in a couple seconds. 

Caching the cache dir still requires yarn to resolve things and map things out, and still takes time to do the install. To me that feels like the cache didn't work, but we all know that it does 😅 

In the grand scheme of things, it doesn't really matter for this repo because it's not terribly huge. 

I'll think about it a bit more.

Phillip9587: @Cammisuli Yeah i totally agree with you that caching the node_modules is quicker. But the node_modules directory could be corrupted. By Definition a CI run should always start with a clean state of the repository. By caching the yarn cache dir the node_modules folder is created with a clean state. To speed things up a bit more i could set the `--prefer-offline flag` if you want me to.

Cammisuli: Yea you're right about the ci. I'm just gonna have to ignore the install step and focus on something else when it's being ran 😅

Don't worry about the additional flag

Phillip9587: But the actions/cache@v2 action has a significant performance improvement over v1 ... so at least the caching steps are faster

Cammisuli: yup, very true!

Cammisuli (approved):

PR #1087, opened 2021-06-17
refactor: removed json5 and strip-json-comments from dependencies and used jsonc-parser instead

Phillip9587:
- Updated `jsonc-parser` to 3.0.0
- Removed `strip-json-comments' from dependencies
- Removed `json5` from `apps/vscode/src/package.json`
- Replaced the `readAndParseJson` function by a faster implementation similar to the one at https://github.com/nrwl/nx/blob/4f429e0eab7f9294184166cef202688d434e21a8/packages/tao/src/utils/json.ts#L35

### DIFF
--- a/libs/server/src/lib/utils/utils.ts
+++ b/libs/server/src/lib/utils/utils.ts
@@ -13,7 +13,11 @@ import {
   ItemsWithEnum,
 } from '@nx-console/schema';
 import { Option as CliOption } from '@angular/cli/models/interface';
-import * as stripJsonComments from 'strip-json-comments';
+import {
+  parse as parseJson,
+  printParseErrorCode,
+  ParseError,
+} from 'jsonc-parser';
 
 export interface SchematicDefaults {
   [name: string]: string;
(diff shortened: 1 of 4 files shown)


libs/server/src/lib/utils/utils.ts
```diff
@@ -121,21 +125,36 @@ export function listFiles(dirName: string): string[] {
...
+    const result = parseJson(content, errors);
+
+    if (errors.length > 0) {
+      const { error, offset } = errors[0];
+      throw new Error(
+        `${printParseErrorCode(error)} in JSON at position ${offset}`
+      );
+    }
+
+    return result;
```
Cammisuli, line 156: This is neat, I have to run a few performance tests locally to see how well this performs. The repo I do for this is here if you want to try it yourself as well:
https://github.com/Cammisuli/nx-incremental-large-repo/tree/1425-projects

There's essentially 1425 projects, and we need to make sure that we can parse this fairly quickly to get a list of all the projects, and their targets.

Cammisuli (approved):

Cammisuli:
Thanks for this @Phillip9587!

PR #1099, opened 2021-06-20
chore(repo): remove obsolete vscode postinstall

Phillip9587:
Removed obsolete vscode postinstall script: node ./node_modules/vscode/bin/install from tools/scripts/postinstall.js see: https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode


Cammisuli (approved):
Thanks! I've seen this happen too often, but never got around to fixing it 👍🏼

PR #1102, opened 2021-06-25
feat: support split nx projects

Cammisuli:
## What it does

Adds support for split nx configs. Nx console will use the locally installed `@nrwl/workspace` package to normalize the workspace with all configs.


libs/vscode/nx-workspace/src/lib/get-nx-workspace-package.ts
```diff
@@ -0,0 +1,16 @@
...
+import type * as NxWorkspacePackage from '@nrwl/workspace';
+/**
+ * Get the local installed version of @nrwl/workspace
+ */
+export function getNxWorkspacePackage(): typeof NxWorkspacePackage {
+  const workspacePath = dirname(
+    WorkspaceConfigurationStore.instance.get('nxWorkspaceJsonPath', '')
+  );
+
+  // webpack hacks..
```
sandikbarr, line 12: 😼  nice! bonus!

libs/vscode/nx-workspace/src/lib/verify-workspace.ts
```diff
@@ -18,13 +20,34 @@ export function verifyWorkspace(): {
...
+    // noop - use old way
+  }
+
   try {
-    const workspaceJsonPath = join(workspacePath, 'workspace.json');
-    const angularJsonPath = join(workspacePath, 'angular.json');
     if (fileExistsSync(workspaceJsonPath)) {
       return {
         validWorkspaceJson: true,
+        // TODO(cammisuli): change all instances to use the new version - basically reverse this to the new format
```
sandikbarr, line 45: 👍  good call on keeping this Legacy format for an initial release, def another thing to tackle, but likely all over the place

sandikbarr (approved):
Looking good here. I'll do some testing on it too. Maybe @AgentEnder can help with testing on it, too. Thanks so much for tackling this!

PR #1103, opened 2021-06-28
fix: read and combine both schematics and generators

Cammisuli:
## What it does
Combines both schematics and generators schema's when reading schematics/generators

### DIFF
--- a/libs/server/src/lib/utils/read-schematic-collections.ts
+++ b/libs/server/src/lib/utils/read-schematic-collections.ts
@@ -165,17 +165,15 @@ async function readCollectionSchematics(
   collectionName: string,
   collectionJson: any
 ) {
-  const schematicCollection = {
-    name: collectionName,
-    schematics: [] as Schematic[],
-  };
+  const generators = new Set<Schematic>();
+
   try {
     Object.entries(
-      collectionJson.schematics || collectionJson.generators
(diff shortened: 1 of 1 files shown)


libs/server/src/lib/utils/read-schematic-collections.ts
```diff
@@ -165,17 +165,15 @@ async function readCollectionSchematics(
...
-  const schematicCollection = {
-    name: collectionName,
-    schematics: [] as Schematic[],
-  };
+  const generators = new Set<Schematic>();
+
   try {
     Object.entries(
-      collectionJson.schematics || collectionJson.generators
+      Object.assign({}, collectionJson.schematics, collectionJson.generators)
```
Cammisuli, line 172: I'd like to only do one or the other based on whether the workspace is an Angular workspace or a Nx one.

libs/server/src/lib/utils/read-schematic-collections.ts
```diff
@@ -211,7 +212,12 @@ export async function readSchematicOptions(
       collectionPackageJson.json.generators,
     dirname(collectionPackageJson.path)
   );
-  const schematics =  collectionJson.json.schematics || collectionJson.json.generators;
+  const schematics = Object.assign(
```
Cammisuli, line 215: Same thing here. We should check if the workspace is one or the other, and only read the appropriate schemas.

PR #1128, opened 2021-09-03
feat: Add commands to add applications and libraries directly

Cammisuli:
## What it does
This PR introduces new context menus and commands to generate Applications and Libraries directly (ie. filtered generators)

This is the implementation for the `x-type` schema property (https://github.com/nrwl/nx/pull/6874)


libs/server/src/lib/utils/read-generator-collections.ts
```diff
@@ -155,29 +158,42 @@ async function readCollection(
...
+  const generators = new Set<Generator>();
 
   try {
     Object.entries(
       Object.assign({}, collectionJson.schematics, collectionJson.generators)
-    ).forEach(async ([k, v]: [any, any]) => {
+    ).forEach(([k, v]: [any, any]) => {
       try {
         if (canAdd(k, v)) {
+          let generatorType: GeneratorType;
```
AgentEnder, line 180: Should we include case's for 'app' & 'lib' here as well? This is not something that we currently standardize when thinking about community plugins. As a real example, my plugin nx-dotnet currently uses app / lib instead of application and library.

libs/vscode/nx-run-target-view/src/lib/run-target-tree-item.ts
```diff
@@ -73,7 +74,8 @@ export class RunTargetTreeItem extends TreeItem {
   constructor(
     readonly workspaceJsonPath: string,
     readonly route: WorkspaceRouteTitle,
-    readonly extensionPath: string
+    readonly extensionPath: string,
+    readonly generatorType?: GeneratorType
```
AgentEnder, line 78: Is this being used anywhere? I don't see any refs to it.

libs/vscode/webview/src/lib/get-task-execution-schema.ts
```diff
@@ -190,20 +192,20 @@ function getConfigValuesFromContextMenuUri(
     const appsDir = nxConfig.workspaceLayout?.appsDir ?? 'apps';
     const libsDir = nxConfig.workspaceLayout?.libsDir ?? 'libs';
     if (
-      (appsDir && schematic.name === 'application') ||
-      schematic.name === 'app'
+      (appsDir && generator.name === 'application') ||
```
AgentEnder, line 195: This is an example of a place where we do check if application || app instead of just the one.

Theoretically this shouldn't be necessary, "well behaved" plugins should stick to application and have app as an alias, but nx-dotnet doesn't so I am not sure who else does.

AgentEnder (changes requested):
Few very minor things, otherwise LGTM

libs/vscode/webview/src/lib/get-task-execution-schema.ts
Cammisuli: I was thinking about changing this to use the new `GeneratorType`, but I didn't want to break existing expectations.

AgentEnder: Yeah, this ones not a recommended change, just a shout out to the comment on read-generator-collections

AgentEnder:
I resolved my threads on this, I was looking at part of this incorrectly 🤷. I somehow missed where it read x-type

AgentEnder (approved):

PR #1129, opened 2021-09-03
fix: remove loading item from projects view

### LINKED ISSUE #1126: 17.9.0 Projects directory is not refreshed
Maybe it's a bug.
when i update `Nx-console` 17.8.0 -> 17.9.0
Nx‘s UI Projects directory is not refreshed

截屏2021-08-31 下午10 38 42

截屏2021-08-31 下午10 38 38

Report this error `Cannot read property ‘Loading’ of undefined`

when i downgrade to 17.8.0 , as usual, I can use

I think maybe it's a bug.

System info:
`
Mac os 11.5.2
`

VS Code:
`
 1.59.1 (Universal)
`

Cammisuli:
## What it does
This removes the loading item from the projects view. This was causing some issues with some users where items were trying to be used before all the projects were loaded.

Initially this was added in this extension as a way to show that projects were loading (duh :joy:) because the parser used originally was terribly slow, and the vscode ui froze. With the quicker parser, we don't freeze the UI anymore, and this loading item is not needed. 

Potentially fixes #1126


AgentEnder (approved):
lgtm

PR #1134, opened 2021-09-17
feat: use vscode settings for shell execution

Cammisuli:
## What it does
This simplies creating the `ShellExecution` class to use defaults provided by vscode. 

More information can be found here:
https://code.visualstudio.com/docs/editor/integrated-terminal#_configuring-the-taskdebug-profile

In summary, Nx Console will now use the default terminal profile, OR the profile provided by the `terminal.integrated.automationShell.<platform>` setting`

### DIFF
--- a/libs/vscode/tasks/src/lib/cli-task.ts
+++ b/libs/vscode/tasks/src/lib/cli-task.ts
@@ -1,5 +1,5 @@
 import { CliTaskDefinition } from './cli-task-definition';
-import { Task, TaskGroup, TaskScope } from 'vscode';
+import { ShellExecution, Task, TaskGroup, TaskScope } from 'vscode';
 import { getShellExecutionForConfig } from './shell-execution';
 import { findClosestNg, findClosestNx } from '@nx-console/server';
 import { join } from 'path';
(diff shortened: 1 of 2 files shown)


libs/vscode/tasks/src/lib/shell-execution.ts
```diff
@@ -14,46 +14,8 @@ export interface ShellConfig {
...
-let bashPath: string;
-function getUnixShellExecution(config: ShellConfig): ShellExecution {
-  if (!bashPath) {
-    try {
-      bashPath = execSync('which bash').toString().trim() || '/bin/bash';
-    } catch {
-      bashPath = '/bin/bash'; // Default to where bash is usually installed.
-    }
-  }
   return new ShellExecution(config.displayCommand, {
```
AgentEnder, line 17: Much Much Simpler than before! 👍

libs/vscode/tasks/src/lib/cli-task.ts
```diff
@@ -1,5 +1,5 @@
 import { CliTaskDefinition } from './cli-task-definition';
-import { Task, TaskGroup, TaskScope } from 'vscode';
+import { ShellExecution, Task, TaskGroup, TaskScope } from 'vscode';
```
AgentEnder, line 2: Is ShellExecution used somewhere here? I don't see a reference within this file. 

Maybe it has side effects?

AgentEnder (approved):
LGTM

PR #1136, opened 2021-09-18
style: migrate division syntax

peterblazejewicz:
Minor fix, removes warnings. Instead of using `math.div`, uses multiplier.
https://sass-lang.com/documentation/breaking-changes/slash-div

Thanks!


Cammisuli (approved):
Thanks for this! I always remember seeing that warning come up but always forget to investigate it 😅

PR #1143, opened 2021-09-25
fix: update default branch text to "main"

mruffieux:
## Motivation?
When creating a new workspace, `nx` init a git repository. Since `git` default branch name policy has changed from `master` to `main`. This lead commands such as `nx affected:build` to crash after a fresh install. All people new to nx will eventually face this issue.

## What's been done?
Replacing `master` with `main` where needed.

## Breaking changes?
If running `nx migrate` replaces some of the updated files, then the commands will break for projects having `master` as their default base. The solution is to use the `--base` flag on such commands.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -1,14 +1,16 @@
 {
   "implicitDependencies": {
-    "angular.json": "*",
-    "package.json": "*",
-    "tsconfig.base.json": "*",
+    "package.json": {
+      "devDependencies": "*",
+      "dependencies": "*"
+    },
     ".eslintrc.json": "*"
   },
   "affected": {
     "defaultBase": "master"
   },
   "npmScope": "nx-console",
+  "useDaemonProcess": true,
   "tasksRunnerOptions": {
     "default": {
       "runner": "@nrwl/nx-cloud",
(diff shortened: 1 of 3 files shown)


nx.json
```diff
@@ -6,7 +6,7 @@
     ".eslintrc.json": "*"
   },
   "affected": {
-    "defaultBase": "master"
+    "defaultBase": "main"
```
AgentEnder, line 9: This affects _this_ repo, not projects using nx console. Since this repo still has master for main branch changing this would cause commands ran in _this_ repo to break, without changing the end user experience.

AgentEnder (changes requested):
This should wait until nrwl/nx#7111 is merged and released. Current default behavior with `create-nx-workspace` is to have a default branch of master (and I'd argue that most Nx workspaces still use master since the change is somewhat recent)

AgentEnder:
This is definitely something we will want to merge when the time is right though! Thanks for the PR 😄

nx.json
mruffieux: Oh I didn't know that 😓 Thanks for the good look @AgentEnder o7

Cammisuli:
I appreciate the work in this. It's definitely something on the radar to do :D

AgentEnder (dismissed):
If you revert the changes in package-scripts.js, and nx.json, this PR should be good to merge. The `master` -> `main` change in the main nx repo is complete and should handle most concerns.

Cammisuli:
I'm going to do some cleanup and get this one merged in.

PR #1160, opened 2021-10-16
perf: asynchronously retrieve all collection info (schematics, generators, builders and executors)

Cammisuli:


Cammisuli:
I don't know if I'd like reading through the whole node_modules. That always seemed kind of overkill. I always wanted to streamline this to only read the dependencies that are actually listed in the package.json.

Adding a flag to keep the old behaviour doesn't feel right for me. 


And the reason why this could potentially be a breaking change is if someone was using schematics/generators that they didn't explicitly add to their dependencies. To me, that's not the correct way and those dependencies should be listed. 

I don't know yet, I'd like to gather some more opinions and see what people think.

Thanks for your feedback :smile:

PR #1184, opened 2021-12-03
feat: use nx utils to read all configuration files

Cammisuli:
## What it does
This PR removes all direct reads to the workspace/project/angular/nx.json files and uses the utils provided by Nx. This allows Nx Console to always know what configurations are available without having to worry about workspace changes.


apps/vscode/src/main.ts
```diff
@@ -116,6 +117,7 @@ export async function activate(c: ExtensionContext) {
     const vscodeWorkspacePath =
       workspace.workspaceFolders && workspace.workspaceFolders[0].uri.fsPath;
 
+    debugger;
```
AgentEnder, line 120: Probably remove?

libs/server/src/lib/utils/utils.ts
```diff
@@ -352,7 +355,9 @@ function renameProperty(obj: any, from: string, to: string) {
   delete obj[from];
 }
 
-export function toWorkspaceFormat(w: any): WorkspaceJsonConfiguration {
+export function toWorkspaceFormat(
```
AgentEnder, line 358: This function could reuse `toNewFormat` from @nrwl/tao

AgentEnder: These linked lines are probably not needed anymore since we use the NX helpers to read config

https://github.com/Cammisuli/nx-console/blob/use-nx-utils-for-configs/libs/server/src/lib/utils/utils.ts#L380-L382

Cammisuli:
Need to finish #1057 before this can complete. The latest versions of angular cli is not as backwards compatible as nx, so we need to completely remove our dependency on ng cli

PR #1186, opened 2021-12-04
feat: provide json schema for nx.json

puku0x:
This PR adds a JSON schema to provide intellisense when configuring `nx.json`.

image

## Related Issue(s)
https://github.com/nrwl/nx/pull/7993


puku0x:
👀
image

Cammisuli (changes requested):
The workspace and project.json were a little more involved because they're auto generated based on whatever workspace Nx Console is opened in. Since the schema for nx.json doesn't need to be dynamic, we don't have to do the whole schema generation. 

Instead we can just place a json file within the project and link that within the package.json in the app directory. 

Add a new object here:
https://github.com/nrwl/nx-console/blob/53b09eb361845498234a6556bc296ae302660b02/apps/vscode/src/package.json#L770-L783
with 
```
{
  "fileMatch": "nx.json",
  "url": "./nx-schema.json"
},
```
Then create a file beside the package.json called `nx-schema.json` and put the string you have in `createJsonSchema` within that json file. 

And that should be it 😄 


Let me know if you have any questions.

puku0x:
Thank you for your advice! I reverted the generator and added `nx-schema.json` into the directory.

Cammisuli (approved):
Thank you for creating the schema for this!

PR #1208, opened 2021-12-28
fix: handle absence of "properties" in schema

dmitry-stepanenko:
I've encountered an issue with some old repository. It contained a schematics, that didn't have `properties` field specified in `schema.json`. Schematics itself was working properly, but `nx-console` plugin failed to even launch. This PR adds handling of a case there's no `properties` field in schema.
image

@vsavkin would it be reasonable to also make `properties` field optional in the `Schema` interface in Nx repo [here](https://github.com/nrwl/nx/blob/master/packages/tao/src/shared/params.ts#L51)?

### DIFF
--- a/libs/server/src/lib/utils/utils.spec.ts
+++ b/libs/server/src/lib/utils/utils.spec.ts
@@ -24,6 +24,12 @@ describe('utils', () => {
       return r;
     };
 
+    it('should work with schema without any properties', async () => {
+      // @ts-expect-error absence of required property "properties" is needed to test failure resistance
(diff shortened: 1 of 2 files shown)


Cammisuli:
Thanks for the PR! Just out of curiosity, what version was that old workspace?

dmitry-stepanenko:
> Thanks for the PR! Just out of curiosity, what version was that old workspace?

@Cammisuli that repo was using Angular and nx v11, but the issue is easily reproducible on latest versions

PR #1209, opened 2022-01-07
fix: support workspace-generators and workspace-schematics

### LINKED ISSUE #1187: Fails to run custom schematics
Nx version: v10
Nx Console version: v17.12.4

In Nx Workspace (v10) where we have custom schematics (instead of custom generators), invoking the custom schematics via Nx Console fails because Nx Console tries to run: `nx workspace-generator generator-name` instead of `nx workspace-schematic schematic-name`

Cammisuli:
## What it does
Fixes running workspace schematics if the workspace version supports that instead of workspace generators

Fixes #1187


ZackDeRose (approved):
LGTM

PR #1215, opened 2022-01-15
fix: fixed CI Node Setup and Caching

Phillip9587:
The `setup` action should use `actions/setup-node` **v2**. It supports caching internally so there is no need for the extra `actions/cache` step. 

Also the Agents are currently running the `actions/setup-node` step twice.


Cammisuli:
Oh nice, I didn't even bother checking for a new version of the setup node. Thank you!

PR #1220, opened 2022-01-18
fix: inline copy-to-clipboard icon

Phillip9587:


Phillip9587:
@Cammisuli Nx Console uses only one icon of the whole codicons: 
https://github.com/nrwl/nx-console/blob/e99e226df146f60b2c4b995c2ddc1307f15b003a/libs/vscode-ui/feature-task-execution-form/src/lib/task-execution-form.component.html#L71 

But the whole font is included. Maybe it should be considered to either use more icons or to inline the one icon.

Cammisuli:
I haven't looked into inlining the icon, but I suspect it might be easier just to copy the svg or something.

Phillip9587:
@Cammisuli I inlined the SVG Icon and removed the deprecated package

PR #1227, opened 2022-01-21
chore: use sass modules

Phillip9587:
Sass discourages the use of the `@import` At-Rule (https://sass-lang.com/documentation/at-rules/import). 

The Sass Module system and `@use` should be used instead.


Cammisuli:
Looks good, thanks again!

PR #1233, opened 2022-01-26
chore: remove shelljs and shx

Phillip9587:
- removed `shelljs` and `shx`
- refactored `vscode-yarn.js` to use `child_process.exexSync()` instead of `shelljs.exec()`
- refactored package-scripts `clean` task to use `rimraf` instead of `shx` 
- removed leftover  package-scripts task `shx rm -rf ${join('dist', 'apps', 'vscode', '**', '*-es5.js')}`


Cammisuli:
Thanks for the clean up as usual! 😄

PR #1240, opened 2022-01-28
fix: run pnp.setup to patch require

Cammisuli:
## What it does
This calls `pnp.setup()` so that it can patch the node `require` so that we can load the `@nrwl/workspace` package from yarn 2+


ZackDeRose (approved):
lgtm

PR #1254, opened 2022-02-25
feat: add prompts to add nx to angular cli

Cammisuli:


libs/vscode/nx-conversion/src/lib/vscode-nx-conversion.spec.ts
```diff
@@ -0,0 +1,7 @@
+import { vscodeNxConversion } from './vscode-nx-conversion';
+
+describe('vscodeNxConversion', () => {
+  it('should work', () => {
+    expect(vscodeNxConversion()).toEqual('vscode-nx-conversion');
+  });
+});
```
ZackDeRose, line 7: Want to remove the spec for now?

libs/vscode/nx-conversion/src/lib/vscode-nx-conversion.ts
```diff
@@ -0,0 +1,139 @@
...
+import {
+  commands,
+  ExtensionContext,
+  Task,
+  tasks,
+  TaskScope,
+  window,
+} from 'vscode';
+
+const PROMPT_COUNT = 1;
```
ZackDeRose, line 14: Think we wanna bump this back up to 5 right?

libs/vscode/nx-conversion/src/lib/vscode-nx-conversion.ts
```diff
@@ -0,0 +1,139 @@
...
+const PROMPT_MSG = 'Convert to Nx?';
```
ZackDeRose, line 15: I was thinking for more along the lines of "Would you like to upgrade the Angular CLI with Nx (making your `ng` commands up to 10 times faster)?"

package.json
```diff
@@ -26,7 +26,8 @@
     "@monodon/typescript-nx-imports-plugin": "0.2.0",
     "@yarnpkg/fslib": "2.6.1-rc.5",
     "@yarnpkg/libzip": "2.2.3-rc.5",
-    "jsonc-parser": "^3.0.0"
+    "jsonc-parser": "^3.0.0",
+    "tslib": "^2.0.0"
```
ZackDeRose, line 30: was this intended to be added?

ZackDeRose (approved):
Couple small questions - other than that lgtm

libs/vscode/nx-conversion/src/lib/vscode-nx-conversion.ts
Cammisuli: Yup, we'll need to think of a good tag line

libs/vscode/nx-conversion/src/lib/vscode-nx-conversion.ts
ZackDeRose:
```suggestion
const PROMPT_COUNT = 5;
```

apps/vscode/src/main.ts
```diff
@@ -53,6 +52,7 @@ import {
   ProjectJsonSchema,
 } from '@nx-console/vscode/json-schema';
 import { enableTypeScriptPlugin } from '@nx-console/typescript-plugin';
+import { NxConversion } from '@nx-console/vscode/nx-conversion';
```
ZackDeRose, line 55: Looks like lint error here

ZackDeRose: Looks like just an error in my IDE

PR #1256, opened 2022-03-04
feat: Nx Move and Remove Project in Context Menu

ZackDeRose:
Hey @Cammisuli - wanted to get your thoughts on this approach before going too far in.

You can see the code for yourself, but the idea here is:

- we'd add the move VsCode command
- we determine the appropriate move generator (given the URI [stubbed out for now - always does `workspace:move`])
- add additional optional parameters to allow this move generator to 'drill down' our existing API

I'm not a fan of the amount of "drilling" this approach takes, but I think it is the minimally invasive approach given our current state. (eventually it'd be nice to rework some of our APIs like we've discussed)


apps/vscode/src/package.json
```diff
@@ -74,6 +74,16 @@
           "command": "nx.run.fileexplorer",
           "group": "2_workspace"
         },
+        {
+          "when": "isNxWorkspace && config.nxConsole.enableGenerateFromContextMenu",
+          "command": "nx.move.fileexplorer",
+          "group": "2_workspace"
+        },
+        {
+          "when": "!isNxWorkspace && config.nxConsole.enableGenerateFromContextMenu",
```
Cammisuli, line 83: I think it's better to use `isAngularWorkspace` here. 
```suggestion
          "when": "isAngularWorkspace && config.nxConsole.enableGenerateFromContextMenu",
```

libs/vscode/tasks/src/lib/cli-task-commands.ts
```diff
@@ -67,6 +75,20 @@ export function registerCliTaskCommands(
       selectCliCommandAndPromptForFlags('run', await getCliProjectFromUri(uri))
     );
 
+    commands.registerCommand(`${cli}.move.fileexplorer`, async (uri: Uri) => {
+      const getCorrectMoveGenerator = (uri: Uri) => '@nrwl/workspace:move';
```
Cammisuli, line 79: We could probably find out the minimum version of Nx that introduced this generator and have that logic here. We have this function here that could be exported and used outside that project:
https://github.com/nrwl/nx-console/blob/16e315521dbd90de3040abc71c28826fba44af8e/libs/vscode/nx-workspace/src/lib/nx-version.ts#L7

Cammisuli (commented):
I just took a quick look, and I see what you mean about the drilling down 😅 that WebView function is doing too much.

Either way, there's a few comments I have as well 😄

libs/vscode/tasks/src/lib/cli-task-commands.ts
ZackDeRose: Oooh, hadn't considered that. Good point.

For versions younger than we had the move/remove generators, we should probably remove this option from the context menu, right?

ZackDeRose: Similar note: for when the user right-clicks on a node in the root, my solution would open the webview for the `@nrwl/workspace:move` generator; without a project set.

I think this is problematic, since we *should* be using the `@nrwl/angular:move` generator if the project is angular :\

ZackDeRose:
@Cammisuli updates:

+ Added logic for getting the correct generator (choose angular/move if installed, otherwise workspace/move) [comments added for future problems with this]
+ Removed this context menu item for Angular CLI workspaces (move and remove only work for nx projects as I understand it)
+ Added a version check to not add it for older nx versions.
+ Left the "Nx Move...." option in the context menu even when not clicking on a project (this will open the generator webview WITHOUT a project set. I think this is probably the best solution, but I could see a valid argument for removing it unless you click on a project as well.)

Taking off the WIP, lmk if anything looks like it's missing!

AgentEnder:
In regards to the @nrwl/angular:move vs @nrwl/workspace:move point, there's also the case of community plugin projects which may require their move generator. e.g., I'm about to add a move generator to nx-dotnet to handle updating .csproj file paths in addition to moving the files (nx-dotnet/nx-dotnet#456).

Do you think we could check for all installed plugins that have a generator called move, and prompt the user to choose between them?

Cammisuli (approved):

PR #1269, opened 2022-04-08
fix: sort properties by required

### LINKED ISSUE #1265: Sort Form Fields by required
## Description
When using the UI to generate, the fields should be sorted by required, after `$default` arguments. 
image

## Motivation
This helps with a visual grep to see which fields are required to actually run the generator. 

## Suggested Implementation

modify sort here:
https://github.com/nrwl/nx-console/blob/6d060e3068ea0f521fb9efc80cd974d14847575a/libs/server/src/lib/utils/utils.ts#L217-L246

Cammisuli:
## What it does
This makes sure that required properties are sorted near the top of the list (there are still some properties that are more important)

Fixes #1265


ZackDeRose (approved):
LGTM

PR #1274, opened 2022-04-22
feat: list local plugins

### LINKED ISSUE #1271: Add locally available plugin generators
## Description

Now that <https://github.com/nrwl/nx/pull/9116> has been released, it would be an excellent enhancement for nx-console to be able to select locally available generators.

## Motivation

Ease of discoverability and alignment with the `nx` CLI.

## Suggested Implementation

When selecting "generate" from nx-console, the latest generators in all available local plugins should appear and be selectable. nx-console should also be able to detect when new plugins are initialized and present an updated list.

## Alternate Implementations

Unsure

Cammisuli:
## What it does
This enables support for local plugins. They will now appear where other generators and builders appear 😄 

Closes #1271


apps/vscode/src/main.ts
```diff
@@ -234,7 +235,7 @@ async function setWorkspace(workspacePath: string) {
     WorkspaceConfigurationStore.instance.set('nxWorkspacePath', workspacePath);
   }
 
-  await setApplicationAndLibraryContext(workspacePath);
+  setApplicationAndLibraryContext(workspacePath);
```
Cammisuli, line 238: I removed the `await` because I want this to be "fired and forgotten"

ZackDeRose (approved):
LGTM

PR #1282, opened 2022-05-13
chore: update analytics/telemetry 

Cammisuli:
We need to update analytics as Universal Analytics is going away next year (July 2023). 

We need to migrate to GA v4, or something else. Right now, there's no packages/tools available for node that post to GAv4, everything is still universal-analytics. Will put this on hold until I have more time to actually write something myself, or someone creates a package. 

# TODO
use this library: https://github.com/DavidWells/analytics/tree/master/packages/analytics-plugin-google-analytics


MaxKless:
I'm closing this as we're going to redo telemetry ourselves soon.

PR #1300, opened 2022-06-30
feat: handle ts paths imports with asterix

moczix:
`    "paths": {
      "@hehe-nx/blabla/*": ["libs/blabla/src/lib/*"],
      "@hehe-nx/bobo/roro/coco": ["libs/bobo/roro/coco/src/index.ts"]
    }`

consider this paths, currently we only handle  **"@hehe-nx/bobo/roro/coco": ["libs/bobo/roro/coco/src/index.ts"]** 
but alot of users have paths like **"@hehe-nx/blabla/*": ["libs/blabla/src/lib/*"],** so i added solution to resolved this files to have imports autocomplete


moczix:
i closed this because its has performance issue in big project

PR #1309, opened 2022-07-20
feat: add an allowlist and a blocklist for generators

### LINKED ISSUE #1272: Add ability to exclude and/or include the list of generators available
## Description

Add the ability to exclude and/or include the list of generators available

## Motivation

Currently when opening the list of Generators, it simply shows any and every generator available to you. This is great for power users, but for orgs where maybe only the custom workspace-generators are needed, it would be nice to limit what is shown in that list to make it easier to target using the correct ones.

## Suggested Implementation

Add a configuration setting that can be passed a list of generators to explicitly show

## Alternate Implementations

Or in addition to that and/or instead of that, add a configuration setting where you can pass a list of generators to hide from that list.

Cammisuli:
## What it does
Adds new settings to enable filtering of the generator list within Nx Console. 

These settings are "nxConsole.enableGeneratorFilters", "nxConsole.generatorAllowlist", "nxConsole.generatorBlocklist".

The allow and block lists can be combined, where the allow list takes priority, then the block list is applied to the remaining items. 

Fixes #1272

### DIFF
--- a/apps/vscode/src/package.json
+++ b/apps/vscode/src/package.json
@@ -676,11 +676,6 @@
           "default": true,
           "description": "Help make Nx Console better by sending anonymous usage data to the Nx Console team."
         },
-        "nxConsole.useNVM": {
-          "type": "boolean",
-          "default": false,
-          "description": "Runs tasks using Node Version Manager"
-        },
         "nxConsole.enableGenerateFromContextMenu": {
           "type": "boolean",
           "default": true,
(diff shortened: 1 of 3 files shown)


jared-christensen:
Is there docs for this, I would like to use this.

jmroon:
I would also like to use this. Is this only relevant to the vscode extension? Where do I set these settings?

jared-christensen:
> I would also like to use this. Is this only relevant to the vscode extension? Where do I set these settings?

settings_json_—_dice-web-platform_—_Dice_Web_Platform

PR #1331, opened 2022-09-06
feat: add help & feedback view

MaxKless:
We want to make it easier for users to access relevant help and information as well as encourage them to provide feedback.

A new view with links to docs & github is introduced.


Cammisuli (approved):
I updated the icon to use the Nx Console one rather than the CLI one 🙂

PR #1336, opened 2022-09-07
docs: introduce architecture.md

MaxKless:


MaxKless:
@Cammisuli this is just a first draft, would be thankful for some feedback on what else to add or if parts are unnecessary

PR #1354, opened 2022-09-21
chore: update vscode

MaxKless:
@types/vscode now at 1.71.1 (with quickpick seperator api!)


Cammisuli (changes requested):
Can you update the version here too:
https://github.com/nrwl/nx-console/blob/81d8bd1e3c6d80b28b9878d4b797c34924f1e3c4/apps/vscode/src/package.json#L27

Cammisuli (approved):

PR #1375, opened 2022-09-30
feat: add connect-to-nx-cloud command

MaxKless:


libs/vscode/nx-help-and-feedback-view/src/lib/nx-help-and-feedback-provider.ts
```diff
@@ -42,4 +62,50 @@ export class NxHelpAndFeedbackProvider extends AbstractTreeProvider<NxHelpAndFee
       ([title, link, icon]) => new NxHelpAndFeedbackTreeItem(title, link, icon)
     );
   }
+
+  async isConnectedToCloud(): Promise<boolean> {
+    const nxJson = (await readJsonFile(
+      join(getWorkspacePath(), 'nx.json')
+    )) as NxJsonConfiguration;
```
Cammisuli, line 69: Use the `nxWorkspace` function to get this information. Right now the nx specific data isn't exposed, but you can get it from here:
https://github.com/nrwl/nx-console/blob/21c5cc5246c6e46c0e959d4346ef93105972a30b/libs/shared/workspace/src/lib/workspace.ts#L95-L100

and return it somewhere here:
https://github.com/nrwl/nx-console/blob/21c5cc5246c6e46c0e959d4346ef93105972a30b/libs/shared/workspace/src/lib/workspace.ts#L103-L120

libs/vscode/nx-help-and-feedback-view/src/lib/nx-help-and-feedback-provider.ts
```diff
@@ -42,4 +62,50 @@ export class NxHelpAndFeedbackProvider extends AbstractTreeProvider<NxHelpAndFee
...
+            { cwd: getWorkspacePath() }
+          );
+
+          commandProcess.stdout.setEncoding('utf8');
+
+          commandProcess.stdout.on('data', (data) => {
+            if (data.includes('I want faster builds')) {
+              commandProcess.stdin.write('yes\r\n', 'utf8');
+            }
+          });
```
Cammisuli, line 104: I'm fairly certain we can do a `--no-interaction` or some other kind of flag to automatically accept the defaults.

libs/vscode/nx-help-and-feedback-view/src/lib/nx-help-and-feedback-provider.ts
MaxKless: Seems like the info I need is already returned in the `workspace` property...

libs/vscode/nx-help-and-feedback-view/src/lib/nx-help-and-feedback-provider.ts
MaxKless: I couldn't find anything like that...

libs/vscode/nx-help-and-feedback-view/src/lib/nx-help-and-feedback-provider.ts
```diff
@@ -42,4 +111,84 @@ export class NxHelpAndFeedbackProvider extends AbstractTreeProvider<NxHelpAndFee
...
+
+    window.withProgress(
+      {
+        location: ProgressLocation.Notification,
+        title: 'Connecting you to Nx Cloud...',
+      },
+      () => {
+        return new Promise((resolve) => {
+          try {
+            if (lt(coerce(nxVersion) ?? '', '12.0.0')) {
```
Cammisuli, line 146: This is just a note to self: move the version parser in a util function to use for other packages (like lerna)

PR #1390, opened 2022-10-21
feat: project tree view

### LINKED ISSUE #1181: Feature Request - Grouping projects from Nx Console Projects section
First of all, thank you for all the work on nx console! We've been using it for awhile now and love it. One side-effect we've started seeing is with a mono-repo that has hundreds of libs, and 20+ applications, it's becoming harder to find projects from the list. 
It would be really nice to have an option to group these up by top level domain folders instead of a flat list. Something that could follow the folder structure in our solution would be awesome!

NicoVogel:
An implementation for the request #1181.

Changes:
- extract the logic to create the project view into separate classes (one for list and a new one for tree)
- change the code names (functions, variables) to be closer to what it actually is
- separate the view items into different types (folder, project, target) to make the code more readable

I know that there is already a PR for it, but A) I could not simply extend the existing PR to what was required and B) there is no guarantee that even if I extend the other PR, that the person will actually merge my suggestions.

> link to the other PR #1338

Non the less, I want to thank @rickvandermey to get this feature going. Because without his involvement, I would probably not have implement the feature myself.

closes #1338
resolves #1181


libs/vscode/nx-project-view/src/lib/nx-project-tree-view.ts
```diff
@@ -0,0 +1,177 @@
...
+  private createFolderTreeItem(
+    path: string,
+    projects?: [string, ProjectConfiguration][]
+  ) {
+    if (projects && projects.length === 1) {
+      const [project] = projects;
+      return this.createProjectTreeItem(project);
+    }
+
+    const folderName = path.split('/').pop() ?? '';
```
MaxKless, line 87: Have you tested this in different OS? I feel like using a hardcoded `/` is prone to causing errors. I would rather use `path` utility methods to abstract things like splitting paths for us.

libs/vscode/nx-project-view/src/lib/nx-project-tree-view.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    return this.getParentOfTargetItem(element);
+  }
+
+  async getChildren(element?: NxTreeViewItem) {
+    if (!element) {
+      return this.createFolders();
+    }
+    if (element instanceof NxFolderTreeItem) {
+      return this.createFolders(element);
```
MaxKless, line 55: Currently folders have a `Edit Workspace Definition` button which doesn't do anything. Please remove those (but maybe a 'show in explorer' button is useful). 

image

MaxKless (changes requested):
Hey this looks really good! I haven't been able to look at everything yet but some small stuff could be fixed. Good job tho looks great

NicoVogel:
I was also thing of changing the config setting from a boolean to a string (enum). In case that another view would be added.

One that I could think of would be grouped by tag.

libs/vscode/nx-project-view/src/lib/nx-project-tree-view.ts
NicoVogel: Fixed, I used `path.sep` to get the platform specific separator and use it as split value.
https://nodejs.org/api/path.html#path_path_sep

For joning, I switched to `path.join`.
https://nodejs.org/api/path.html#pathjoinpaths

libs/vscode/nx-project-view/src/lib/nx-project-tree-view.ts
NicoVogel: Fixed, for folders it now shows `Reveal in Explorer`

libs/vscode/nx-project-view/src/lib/nx-project-tree-view.ts
NicoVogel: I was just trying to validate it for windows and it does not yet work...
image
I will look into it

libs/vscode/nx-project-view/src/lib/views/nx-project-tree-view.ts
```diff
@@ -0,0 +1,171 @@
...
+    getParent: listView.getParent.bind(listView),
+  };
+}
+
+class TreeView extends BaseView {
+  constructor(cliTaskProvider: CliTaskProvider) {
+    super(cliTaskProvider);
+  }
+
+  async getParent(element: NxTreeViewItem) {
```
MaxKless, line 32: From what I gather, the `getParent` method is not really used to populate the TreeView but only for the `reveal` API: https://code.visualstudio.com/api/references/vscode-api#TreeDataProvider

One of two things would be great:
- Remove the `getParent` code to make it cleaner (lmk if I missed something and there's a good reason for it to be there)
- Add a 'Focus current project in projects view' or similar command that opens the project view to the project corresponding with the open file

libs/vscode/nx-project-view/src/lib/views/nx-project-util.ts
```diff
@@ -0,0 +1,23 @@
+import path = require('node:path');
+
+export function isDefined<T>(val?: T): val is T {
+  return !!val;
+}
+
+export const PathHelper = {
+  dirs(val: string) {
+    return val.split(path.sep);
```
MaxKless, line 9: As discussed, manually splitting is prone to errors. Try to use node.js path functions.

libs/vscode/nx-project-view/src/lib/nx-project-tree-view.ts
NicoVogel: I noticed 2 things:
1. `TreeView.groupByRootPath` does not work under windows
2. `cliTaskProvider.getProjects()` returns different results depending on if the `node_modules` are installed or not.

The first one I will fix in a day or two.

The latter one should generally be fixed in the project, but I will only provide a workaround in this PR (to keep the PR clean).
It should look like:
```ts
// cli-task-provider.ts

export type NodeModulesNotInstalledWorkspaceProjects = {
  [project: string]: string;
};

// ...
async getProjects(
    json?: WorkspaceJsonConfiguration
  ): Promise<WorkspaceProjects | NodeModulesNotInstalledWorkspaceProjects> {
    if (json) {
      return json.projects;
    } else {
      const result = await nxWorkspace(getWorkspacePath(), outputLogger);
      if (!result.validWorkspaceJson || !result.workspace) {
        return {};
      } else {
        return result.workspace.projects;
      }
    }
  }
```

MaxKless (changes requested):
Thanks for your work. I think it's a great solution and will help many. Just a few things to fix :)

NicoVogel:
> Thanks for your work. I think it's a great solution and will help many. Just a few things to fix :)

Just read through the review comments and I agree with your suggestions.

I was also wondering it it makes sense to cache the result of `cliTaskProvider.getProjects()` and only update it in case of a reload 🤔

MaxKless:
Yes you can skip the memoization and extra features for now :) just the bugfix stuff and naming clarity I would like to see

libs/vscode/nx-project-view/src/lib/views/nx-project-tree-view.ts
NicoVogel: Removed

libs/vscode/nx-project-view/src/lib/views/nx-project-base-view.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (!target) {
+      return;
+    }
+
+    const { configurations } = target;
+    if (!configurations) {
+      return;
+    }
+
+    return Promise.all(
```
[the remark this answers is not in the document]
NicoVogel: I missed to remove it, while refactoring the code :)
Removed

libs/vscode/nx-project-view/src/lib/views/nx-project-tree-view.ts
```diff
@@ -0,0 +1,171 @@
...
+      return rootFolders.map(([path]) => this.createFolderTreeItem(path));
+    }
+
+    const subFolders = this.getSubFolders(map, parent.path);
+    return subFolders.map(([path, projects]) =>
+      this.createFolderTreeItem(path, projects)
+    );
+  }
+
+  private createFolderTreeItem(
```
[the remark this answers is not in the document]
NicoVogel: Fixed

libs/vscode/nx-project-view/src/lib/nx-project-tree-view.ts
NicoVogel: Ok, then I will keep my consideration for this minimal. But I have to consider it at least bit, otherwise the tree view will show nothing at all in case that the node_modules are not installed.

If that is fine for you, then I will leave it as is. Otherwise, as said, I would try to keep it as minimal as possible

libs/vscode/nx-project-view/src/lib/views/nx-project-util.ts
NicoVogel: Fixed and validated via unit tests.

libs/vscode/nx-project-view/src/lib/nx-project-tree-view.ts
NicoVogel: I implemented the minimum, as mentioned

Cammisuli: Yes, it's fine if it shows nothing for now, thank you! 🙂

libs/vscode/nx-project-view/src/lib/views/nx-project-tree-view.ts
```diff
@@ -102,22 +94,22 @@ class TreeView extends BaseView {
...
-  private groupByRootPath(projectDefs: {
-    [projectName: string]: ProjectConfiguration;
-  }) {
-    return Object.entries(projectDefs)
+  private groupByRootPath(projectDefs: ProjectDefinition): TreeViewMap {
+    return objectEntries(projectDefs)
       .flatMap((project) => {
         const [projectName, projectDef] = project;
-        const { root } = projectDef;
+        const nodeModulesInstalled = typeof projectDef === 'string';
```
MaxKless, line 101: Can you maybe provide a comment on why `projectDef` being a string means the node_modules are installed and what this means for the projects view? I think it would help both current and future me understand what's going on here.

libs/vscode/nx-project-view/src/lib/views/nx-project-util.ts
```diff
@@ -1,23 +1,88 @@
...
+}[keyof T][];
+
+export function objectEntries<T extends object>(val: T) {
+  return Object.entries(val) as Entries<T>;
+}
+
+export class PathHelper {
+  constructor(private pathApi: typeof path = path) {}
+
+  detailedDirs(val: string) {
```
MaxKless, line 20: This is quite the function... I get the pain of working with paths in node though. Still, would be great if you could provide typings here. 
Just to make it readable.

libs/vscode/nx-project-view/src/lib/views/nx-project-base-view.ts
```diff
@@ -9,34 +8,23 @@ import {
...
+  getProjects(): Promise<ProjectDefinition>;
+}
 
-    return this.createTargetTreeItem(nxProject, [
-      nxTarget.name,
-      projectDef.targets[nxTarget.name],
-    ]);
-  }
+export abstract class BaseView {
+  constructor(protected readonly infoProvider: ProjectInfoProvider) {}
```
MaxKless, line 27: Can you tell me about your motivation behind the `infoProvider` abstraction? Since we still have to inject `CliTaskProvider` in the different strategies, I don't really see the upside...

MaxKless (changes requested):
Alright we're getting there!
Sadly, I've found one major problem with the new tree view: compatibility with angular workspaces. As you can see from the screenshot, the view is just empty currently.
To work on it, you can copy the testworkspace-ng folder from the `vscode-e2e` app and test it out! 
image

I know I've been saying it a lot but thanks again for your work and the super fast iteration times!!

libs/vscode/nx-project-view/src/lib/views/nx-project-tree-view.ts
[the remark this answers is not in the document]
MaxKless: Is it? Still looks the same to me...

libs/vscode/nx-project-view/src/lib/nx-project-tree-view.ts
NicoVogel: I reverted the change, as it is fine to have it empty for now.
This also reduces the complexity of the PR 😁

libs/vscode/nx-project-view/src/lib/views/nx-project-base-view.ts
NicoVogel: The motive was to create an interface that already has the correct typings. But, as this will be addressed separately, I reverted the change.

NicoVogel:
I have a suspicion that the tree view does not work for angular projects which have sub-projects.
So, I will investigate here as well

NicoVogel:
> [...] 
> I know I've been saying it a lot but thanks again for your work and the super fast iteration times!!

Thank you for the nice words :)
But of course, I did not have time to address your feedback right after you wrote that :D

NicoVogel:
> I have a suspicion that the tree view does not work for angular projects which have sub-projects. So, I will investigate here as well

As expected.
It works with a new setup like: https://github.com/saaivs/ng-multi-projects
But for the older ones it does not work, like: https://github.com/JoelViney/angular-multiple-applications-example

NicoVogel:
@MaxKless Do you also have any requirements regarding the git history?
I would squash everything that we did after my initial PR into the `feat: implement project tree view` commit.
Because I don't think that having all the adjustments we did in the PR is of any value for the future.

So, the PR will only have these 4 commits:
image

Cammisuli:
@NicoVogel we squash all prs into master. So whatever is set as the title of this PR will be the commit message that appears on master.

Cammisuli:
So far everything looks good, thank you! I'm just going to do smoke test on windows and if everything runs well there, I'll merge it in. 

Thank you for the contribution and your patience with our feedback 🙂

Cammisuli (approved):
Thanks again!

NicoVogel:
@MaxKless I validated the performance of the TreeView.
I used this blog as a base as I never tested the performance of a vscode extension: https://www.nicoespeon.com/en/2019/11/fix-vscode-extension-performance-issue/

# Actions I did

1. open nx-console in vscode
2. start `yarn watch`
3. start the extension with `F5`
4. open nx-examples (make sure that tree view is used)
5. start profiling
6. open NX Project view
7. navigate into the projects view
8. click on reload button
9. stop profiling

# Measures

## `createRoot`
image

- not measurable how long it takes
- only `getProjects` decides how long it takes

## `createFoldersOrProjectFromFolder`
image

**first measure**

- not measurable how long it takes
- only `getProjects` decides how long it takes

**second measure**

- takes 0.1ms (self)
- `getProjects` takes about half of the total time

## `createTargetsFromProject`
image

- not measurable how long it takes
- only `getProjects` decides how long it takes

> same result 3x

## Reload with tree already open

- `createRoot` -> same as without reload
- `createFoldersOrProjectFromFolder` -> same as first measure
- `createTargetsFromProject` -> takes 0.1ms (self), `getProjects` takes about 2/3 of the total time
- total time (checking the top most function) 3ms
  - there is some idle time before vscode calls again the `getChildren` function
  - It is called 3 times (for each level) and it takes 1ms each time

# What now?

The tree view is more than fast enough!
Even improving the `getProjects` does not make sense, because the total aggregated time of all `getProjects` calls was 0.80ms.

Cammisuli:
The whole of Nx console is about to be quicker with #1402 as well. Basically the longest running operation that plagued console is now handled in the language server, which is in a separate process 🙂

MaxKless:
Awesome @NicoVogel thx for all the effort in this PR. Really high quality contribution. I would give a shoutout on twitter but you don't have one do you? :O

PR #1393, opened 2022-10-25
fix: debounce file watches to limit thrashing

Cammisuli:


apps/vscode/src/main.ts
```diff
@@ -338,34 +337,41 @@ async function registerWorkspaceFileWatcher(
...
+    )(() => {
+      commands.executeCommand(REFRESH_WORKSPACE);
+    }),
+    watchFile(
+      new RelativePattern(
+        workspacePath,
+        `${basePackagesPath}/**/{project,package}.json`
+      )
+    )(() => {
+      commands.executeCommand(REFRESH_WORKSPACE);
```
Cammisuli, line 374: Previously we created at max 3 file watches for each directory, but now it's always going to be one based on the glob.

PR #1402, opened 2022-11-03
feat: move all workspace retrieval to the language server

Cammisuli:


libs/vscode/nx-workspace/src/lib/get-nx-workspace.ts
```diff
@@ -0,0 +1,6 @@
+import { NxWorkspaceRequest } from '@nx-console/language-server/types';
+import { sendRequest } from '@nx-console/vscode/lsp-client';
+
+export function getNxWorkspace(reset?: boolean) {
```
MaxKless, line 4: maybe add an explicit typing here for clarity?

PR #1453, opened 2022-12-27
fix: fall back to nx.app if no cloud url is set in nx.json

MaxKless:


libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
```diff
@@ -668,6 +666,11 @@ export class NxCloudService extends StateBaseService<InternalState> {
     const nxCloudTaskRunner = Object.values(nxConfig.tasksRunnerOptions).find(
       (r) => r.runner == '@nrwl/nx-cloud'
     );
-    return nxCloudTaskRunner?.options?.url;
+
+    // remove trailing slash
+    return (nxCloudTaskRunner?.options?.url ?? 'https://nx.app').replace(
```
StalkAltan, line 671: We can default to `cloud.nx.app` since the user will be forced through a redirect if trying to access `nx.app`

StalkAltan (approved):

PR #1458, opened 2022-12-28
chore: distinguish workspace DTE states

MaxKless:


libs/vscode/nx-cloud-view/src/webview/components/status-labels.ts
```diff
@@ -15,6 +15,17 @@ export class StatusLabels extends LitElement {
...
+      ? 2
+      : this.state?.isUsingCloudRunner
+      ? 1
+      : 0;
+    const dteLabel =
+      dteStatus === 2
+        ? 'DTE enabled'
+        : dteStatus === 1
+        ? 'No distributed tasks executed yet'
+        : '';
```
meeroslav, line 28: Why no title here?
Could be something like "Not using Nx Cloud"

meeroslav: Otherwise looks good

meeroslav (approved):

PR #1465, opened 2023-01-06
docs(nxls): add lsp-zero configuration

nartc:


Cammisuli:
Since `nxls` is available with lsp-config and mason, we can simplify this a lot:

First install nxls with Mason `:MasonInstall nxls`

Then add this config
```lua 
local lsp = require('lsp-zero')
lsp.configure('nxls', {
    force_setup = true, -- this ensures lsp-zero runs lspconfig.nxls.setup {}
})
-- your other configs
lsp.setup()
```

PR #1477, opened 2023-01-16
chore: add document completion for intellij via lsp

Cammisuli:
Adds completion support for nx related files via the lsp for intellij


apps/intellij/src/main/kotlin/dev/nx/console/lsp/NxlsProcess.kt
```diff
@@ -1,4 +1,4 @@
-package dev.nx.console.languageServer
+package dev.nx.console.lsp
```
MaxKless, line 1: I think we should decide on a consisent naming convention. 
Since the LSP app is called nxls, we should probably stick with that here. Or rename everything to `lsp` but I prefer `nxls` since the class is called `NxlsProcess`

MaxKless (approved):

PR #1481, opened 2023-01-17
refactor: move collection retrieval functions to lsp

MaxKless:


apps/nxls/src/main.ts
```diff
@@ -233,6 +239,23 @@ connection.onRequest(NxWorkspaceRequest, async ({ reset }) => {
...
+    options?: GetGeneratorsOptions;
+  }) => {
+    if (!WORKING_PATH) {
+      return new ResponseError(
+        1000,
+        'Unable to get Nx info: no workspace path'
+      );
+    }
+
+    return getGenerators(WORKING_PATH, args.projects, args.options);
```
Cammisuli, line 255: You don't need to pass projects as an arg to the lsp. It can get it by itself with `nxWorkspace`

```suggestion
const {
      workspace: { projects },
    } = await nxWorkspace(WORKING_PATH, lspLogger);

    return getGenerators(WORKING_PATH, projects, args.options);
```

Cammisuli (changes requested):

apps/nxls/src/main.ts
MaxKless: done, i just get the projects inside the readCollections function now. So I don't have to pass it around needlessly

PR #1494, opened 2023-01-26
feat(intellij): include nxls in the intellij bundle

Cammisuli:


apps/intellij/build.gradle.kts
```diff
@@ -77,47 +75,64 @@ qodana {
...
+        dependsOn("buildNxls")
+        from("${rootDir}/node_modules/node-gyp-build") {
+            into("${rootProject.name}/nxls/node_modules/node-gyp-build")
+        }
+        from("${rootDir}/node_modules/@parcel/watcher") {
+            into("${rootProject.name}/nxls/node_modules/@parcel/watcher")
+        }
+        from(nxlsRoot) {
+            include("**/*.js")
+            into("${rootProject.name}/nxls")
```
Cammisuli, line 95: This is the important bit to include nxls in the intellij zip

apps/vscode/project.json
```diff
@@ -139,6 +139,7 @@
       "dependsOn": ["build", "^build"],
       "options": {
         "commands": [
+          "node ./tools/scripts/copy-nxls-to-vscode.js",
```
MaxKless, line 142: This works but I think it's pretty confusing.
Currently, `nxls:build` and `nxls:watch` will output their bundles to different locations. 
That works because we don't currently have a watch mode for intellij.
But I think it's pretty confusing and I would expect there to be a `dist/apps/nxls` folder even if I'm just spinning up a local instance of VSCode.

I'm not sure if there's a good solution for it since Nx is still a bit wonky with long-running tasks but lmk what you think.

Cammisuli: Nxls is special. We have 3 ways of using it. 
* Within VSCode
* Within Intellij
* Published to NPM

Using `yarn watch` is specifically made for vscode. And that includes watching changes for nxls to make sure it appears in the right spot for vscode to reload the changes properly. If we didnt put the watch results in the vscode folder, then we would need to have a way to copy the asset to the vscode folder. And there's no way to do that atm with the current watch we're using. (We'd need to write a webpack plugin or something to copy after emit). 

Then, `nx run nxls:build` is used for the packaging of the vscode extension, and the usage for intellij. This uses esbuild to make sure that the output is a self contained javascript file that can be executed with `node` directly. The `nx run vscode:package` command makes sure that self contained nxls is moved to the correct spot. The same thing happens with intellij's `prepareSandbox`, we copy the self contained nxls to a spot where intellij can read it. 

And last, we publish to npm. Which goes to the `dist/package/` folder. 

I know it's a lot for the language server, but because we have to use it now in both vscode and intellij, we need to output it into it's own folder for plugin packaging.

PR #1501, opened 2023-01-31
fix: add missing package name and version to migrate command

### LINKED ISSUE #1484: Clicking 'migrate' under 'Common Nx Commands' errors for lack of package name and version
## Current Behavior

Clicking _nx migrate_ does nothing. The command requires a package namd and version, which aren't provided when clicking in the GUI.

image

## Expected Behavior

If `migrate` is expected to take in CLI flags, the console should either:

* Expose some way to provide them
* Not surface migrate in the first place
 
## Steps to Reproduce

1. Install the Nx Console in VS Code
2. Open a repo (I was on https://github.com/typescript-eslint/typescript-eslint)
3. Click _Common Nx Commands_ > _migrate_ 

### Failure Logs / Images / Videos

```plaintext
 *  Executing task: yarn nx migrate 

yarn run v1.22.19
$ /Users/josh/repos/typescript-eslint/node_modules/.bin/nx migrate

 >  NX   Provide the correct package name and version. E.g., my-package@9.0.0.

   Pass --verbose to see the stacktrace.

Command failed: /var/folders/1_/pn8_jg_93cqdnkyswmmr5sxm0000gn/T/tmp-56163-T3ZeQVd4zR4N/node_modules/.bin/nx _migrate 
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Environment

- Nx Console version: v17.28.2
- VSCode version: Version: 1.74.3 (Universal)

```plaintext
$ npx nx report

 >  NX   Report complete - copy this into the issue template

   Node : 18.7.0
   OS   : darwin arm64
   yarn : 1.22.19
   
   nx : 15.3.2
   @nrwl/angular : Not Found
   @nrwl/cypress : Not Found
   @nrwl/detox : Not Found
   @nrwl/devkit : 15.3.2
   @nrwl/esbuild : Not Found
   @nrwl/eslint-plugin-nx : Not Found
   @nrwl/expo : Not Found
   @nrwl/express : Not Found
   @nrwl/jest : 15.3.2
   @nrwl/js : Not Found
   @nrwl/linter : 15.3.2
   @nrwl/nest : Not Found
   @nrwl/next : Not Found
   @nrwl/node : Not Found
   @nrwl/nx-cloud : 15.0.2
   @nrwl/nx-plugin : Not Found
   @nrwl/react : Not Found
   @nrwl/react-native : Not Found
   @nrwl/rollup : Not Found
   @nrwl/schematics : Not Found
   @nrwl/storybook : Not Found
   @nrwl/web : Not Found
   @nrwl/webpack : Not Found
   @nrwl/workspace : 15.3.2
   typescript : 4.9.3
   ---------------------------------------
   Local workspace plugins:
   ---------------------------------------
   Community plugins:
         lerna: 6.4.0
```

cc @AgentEnder, who was helping me with Nx things when we bumped into this. 😄

dzhavat:
Show prompts for selecting a package name and version when running the `migrate` command.

https://user-images.githubusercontent.com/1096332/215871320-fd9607a3-7562-456d-a7d1-dcf8e6b2cc70.mp4

**Note**: the new `dependency-versioning.ts` file is a copy of the same file found in `\libs\vscode\add-dependency\src\lib\dependency-versioning.ts`. It's part of the PR only for starting a discussion about figuring out where to place it so it's available for both libraries. Importing `dependency-versioning.ts` to the `tasks` library causes circular dependency error.

Closes #1484


dzhavat:
> `dependency-versioning.ts` imports from vscode so it has to go into the `libs/vscode/...` folder ... 

Good point. Well, `libs/vscode/utils` seems like the next best place then. Will move it there.

dzhavat:
The function is now moved to `utils` lib. The PR should be ready for review :)

libs/vscode/tasks/src/lib/nx-task-commands.ts
```diff
@@ -341,15 +354,58 @@ async function promptForList() {
...
+
+  const { workspacePath } = await getNxWorkspace();
+
+  const packageJson = await readAndParseJson(
+    join(workspacePath, 'package.json')
   );
-  tasks.executeTask(task);
+
+  if (
+    packageJson.dependencies == undefined ||
```
MaxKless, line 366: currently this would exit even if `packageJson.dependencies` is defined because of the `||`. 
That's probably the wrong approach since only one of them can be defined and it's perfectly valid.

libs/vscode/tasks/src/lib/nx-task-commands.ts
```diff
@@ -341,15 +354,58 @@ async function promptForList() {
...
+    packageJson.devDependencies == undefined
+  ) {
+    return;
+  }
+
+  const dependencyToMigrate = await new Promise<string | undefined>(
```
MaxKless, line 372: Couldn't this be a simple `showQuickPick` call? I don't see a reason to manually construct a quick pick here

libs/vscode/tasks/src/lib/nx-task-commands.ts
```diff
@@ -341,15 +354,58 @@ async function promptForList() {
...
+
+  if (dependencyToMigrate === undefined) {
+    return;
+  }
+
+  const depVersioningInfo = await resolveDependencyVersioning(
+    dependencyToMigrate
+  );
+
+  if (depVersioningInfo !== undefined) {
```
MaxKless, line 396: i mean this is suuuuuuuuper nit picky but if you were to keep with the above coding style, it would be more consistent like this :P 
```suggestion
  if (depVersioningInfo !== undefined) {
  	return
 }
```

libs/vscode/tasks/src/lib/nx-task-commands.ts
```diff
@@ -341,15 +354,58 @@ async function promptForList() {
...
+  );
+
+  if (dependencyToMigrate === undefined) {
+    return;
+  }
+
+  const depVersioningInfo = await resolveDependencyVersioning(
+    dependencyToMigrate
+  );
+
```
MaxKless, line 395: This was the case before you but how about a quick check here or in `dependency-versioning.ts` about whether the dep we're trying to install exists? 
Because if we type in garbage in the dependency quickpick right now, nothing happens. Probably a simple error notification 'The dependency ${dependency} you've selected doesn't exist' or something.

libs/vscode/tasks/src/lib/nx-task-commands.ts
dzhavat: You're right. I think I'll go with assigning an empty object in case one of the properties is `undefined`.

libs/vscode/tasks/src/lib/nx-task-commands.ts
dzhavat: Just tested it and it works. Thanks for the tip. Will go with it.

libs/vscode/tasks/src/lib/nx-task-commands.ts
dzhavat: Used it only because other functions in the same file use similar approach :)

libs/vscode/tasks/src/lib/nx-task-commands.ts
dzhavat: After doing some quick tests, it seems that `showQuickPick` either returns `QuickPickItem` whenever an item from the list is selected, or `undefined` whenever the list is dismissed without selecting anything. It's not possible to submit random text. It must be an item from the list. I've added a notification in case the list is dismissed.

As for selecting a version, dismissing the list doesn't return anything from `resolveDependencyVersioning`. So `promptForVersion` should probably be tweaked to account for that case.

MaxKless: Ah perfect! With the earlier implementation it was still possible but using `showQuickPick` fixed it!

I've removed the info message on cancel because it doesn't fit with our other flows when we cancel quickpicks 
:)

MaxKless (approved):

MaxKless:
thanks so much for your contribution, @dzhavat !!

Issue #1527, opened 2023-02-13
Support wildcards in `nxConsole.generatorBlocklist` and `nxConsole.generatorAllowlist`


Tungsten78:
## Description

Currently, the array values must be of the value `[package]:[generator]` such as `@nrwl/workspace:library`.
From #1309.

In order to curate a refined list for a team this means potentially dozens of explicit callouts.   Not to mention new generators may sneak in when unwanted.

Would like the ability to block all generators within a package, or a subset of generators possibly via wildcard.

## Motivation

Nx generators are ubiquitous.  Being able to limit the options benefits team adoption and lowers cognitive load.

## Suggested Implementation

Match all within a package by omitting the generator.   `@nrwl/workspace`
Or if explicit is preferred via wildcard `@nrwl/workspace:*`.

Optionally by pattern could be useful `@nrwl/cypress:cypress*`

MaxKless:
Yeah this is a great point, I agree. Would you be willing to contribute this? Because it shouldn't be too hard to implement.
Either way, it's going on my list :)

Tungsten78:
I would be happy to - I'll see if I can allocate cycles this week to put it together.

PR #1531, opened 2023-02-15
feat: add wildcard support to generatorAllowlist and generatorBlocklist

Tungsten78:
addresses #1527


MaxKless (approved):
Wow, you're fast! 
Thanks so much for the contribution, works like a charm :)

PR #1534, opened 2023-02-16
feat: enable selecting move generator instead of defaulting to @nrwl/angular

MaxKless:
- refactor: move nx version retrieval to nxls
- feat(vscode): enable choosing (re)move generator from file explorer
- feat: provide wildcard pattern setting for (re)move generators
- refactor: revert selectgenerator changes

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,5 +10,9 @@
     "add-dev-dependency",
     "migrate"
   ],
+  "nxConsole.moveGeneratorPatterns": {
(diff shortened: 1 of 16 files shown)


Cammisuli (changes requested):
Do you think we can split the nxls changes so that the change logs will be cleaner for intellij? 

We don't have this feature in that extension yet and having this in the change log will be confusing.

MaxKless:
closed in favor of two new PRs

PR #1537, opened 2023-02-16
feat(intellij): provide file links for created/updated files

iguissouma:
This PR will provide links for created/updated files on console

Screenshot 2023-02-16 at 22 19 40


MaxKless (approved):
works like a charm, thanks!
Interesting that it automatically only shows links if the files exist!

PR #1548, opened 2023-02-21
fix(intellij): open nx generate ui file on same window

iguissouma:


Cammisuli (approved):
Thank you! Looks good

PR #1549, opened 2023-02-21
fix(intellij): use the local binary for nx generate instead of package managers

Cammisuli:
## What it does
This will use the local install of Nx to run the generate command. This removes the need to check for package managers and just run nx directly.

Should also enable easier support when we move to standalone nx

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/generate_ui/run_generator/RunGeneratorManager.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/generate_ui/run_generator/RunGeneratorManager.kt
@@ -1,5 +1,6 @@
 package dev.nx.console.generate_ui.run_generator
 
+import com.intellij.execution.ExecutionException
 import com.intellij.execution.configurations.GeneralCommandLine
 import com.intellij.execution.executors.DefaultRunExecutor
 import com.intellij.execution.filters.TextConsoleBuilderFactory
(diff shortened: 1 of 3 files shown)


apps/intellij/src/main/kotlin/dev/nx/console/generate_ui/run_generator/RunGeneratorManager.kt
```diff
@@ -48,28 +54,27 @@ class RunGeneratorManager(val project: Project) {
...
-        val pkgManagerParameters =
-            pkgManagerExecCommand.let {
-                if (it.size == 1) emptyList() else it.subList(1, it.size)
-            } + definition
-
         try {
             ApplicationManager.getApplication()
                 .invokeLater(
                     {
+                        val binPath =
```
MaxKless, line 61: I think this should be reworked to also work with [encapsulated nx](https://nx.dev/more-concepts/encapsulated-nx-and-the-wrapper) where there is no `node_modules`

MaxKless (changes requested):

apps/intellij/src/main/kotlin/dev/nx/console/generate_ui/run_generator/RunGeneratorManager.kt
Cammisuli: yea, might as well support it here now

MaxKless (approved):

PR #1552, opened 2023-02-22
feat(intellij): add (re)move project action

MaxKless:
Here's some screenshots showing off the new dialog:

### with comboBox to select one of multiple move generators
image

### fixed generator if there's only one option
image

### autocomplete for projects
image

### dynamic `appsDir` or `libsDir` hint to show what the paths will be relative to (depends on the selected project)
image

### basic validation before running
image

### no destination field when removing
image


bcabanes:
If you are on a "package based" setup meaning you do not necessary have the `/libs` and `/apps` does it work as well?

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxReMoveProjectDialog.kt
```diff
@@ -0,0 +1,197 @@
...
+                        .layout(RowLayout.PARENT_GRID)
+                    row {
+                            label("Destination:")
+                            if (workspaceLayout != null) {
+                                val initialDirHint =
+                                    model.project.let {
+                                        projectOptions?.get(it)?.let { getDestinationDirHint(it) }
+                                    }
+                                        ?: ""
+                                appsOrLibsText = text(initialDirHint).gap(RightGap.SMALL).component
```
Cammisuli, line 123: To always remove the gap
```suggestion
                                initialDirHint?.let {
                                    appsOrLibsText = text(it).gap(RightGap.SMALL).component
                                }
```

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxReMoveProjectDialog.kt
```diff
@@ -0,0 +1,201 @@
...
+        ActionManager.getInstance().getAction(IdeActions.ACTION_CODE_COMPLETION)
+      )} for project completion\n"
+    }
+
+    fun updateDestinationDirHint(projectType: String?) {
+        val appOrLibDirByProjectType = getDestinationDirHint(projectOptions?.get(projectType))
+        appsOrLibsText.let {
+            if (it == null) {
+                return
+            }
```
Cammisuli, line 163: You can remove the check for null by using `?.let`
```suggestion
        appsOrLibsText?.let {
```

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxReMoveProjectDialog.kt
```diff
@@ -0,0 +1,201 @@
...
+        if (workspaceLayout == null) {
+            return ""
+        }
+        if (projectType == "application") {
+            return workspaceLayout.appsDir.let {
+                if (it == null || it == "") {
+                    return ""
+                }
+                "${it}/"
+            }
```
Cammisuli, line 183: We could probably use the `isNullOrEmpty` function to do this check as well.
```suggestion
            return workspaceLayout.appsDir.let {
                if (it.isNullOrEmpty()) {
                    return ""
                }
                "${it}/"
            }
```

Cammisuli (approved):
I have some suggestions but they're not blocking. You can merge this in when you're ready. 🙂

bcabanes (approved):

PR #1553, opened 2023-02-22
chore(generate-ui): remove argument list component

MaxKless:
without the argument list, there's more space for the form! It looks good on smaller screens:
image

For wider resolution, the flag descriptions start to look kind of whacky: 
image

Having the right-aligned button makes no sense anymore imo if the descriptions don't end at the right.


Cammisuli:
We should probably make sure that container has a max width

MaxKless:
I added a max width and centered the form because it now looked too centered on the left. What do you think?
image
image
image

MaxKless:
Updated with bigger input fields on large screens:
image

image
image
image

Cammisuli (approved):

PR #1559, opened 2023-02-22
fix: update nx-console 16x16 icon

Cammisuli:


apps/intellij/src/main/kotlin/dev/nx/console/Icons.kt
```diff
@@ -3,6 +3,6 @@ package dev.nx.console
 import com.intellij.openapi.util.IconLoader
 
 object NxIcons {
-    @JvmField val Action = IconLoader.getIcon("icons/nx-icon.png", javaClass)
-    @JvmField val FileType = IconLoader.getIcon("icons/nx-icon.png", javaClass)
+    @JvmField val Action = IconLoader.getIcon("icons/nx-console_dark.svg", javaClass)
```
iguissouma, line 6: I think we should specify icons without dark suffixe

Screenshot 2023-02-23 at 07 49 20

Cammisuli: Ah damn you're right. I don't know how that slipped through.

Cammisuli: I know who it slipped through, I renamed the file and intellij updated the name here 😅

PR #1573, opened 2023-02-27
fix(intellij): throw explicit error when trying to (re)move and no generators exist

MaxKless:


apps/intellij/src/main/kotlin/dev/nx/console/generate/actions/NxReMoveProjectActions.kt
```diff
@@ -54,6 +54,13 @@ open class NxReMoveProjectActionBase(val mode: String) : AnAction() {
 
         val generators = nxlsService.generators()
         val moveGenerators = generators.map { it.name }.filter { it.contains(Regex(":${mode}$")) }
+
+        if (moveGenerators.isEmpty()) {
+            throw Exception(
+                "No $mode generators found. Make sure the nxls is configured correctly."
```
Cammisuli, line 60:
```suggestion
                "No $mode generators found. Make sure that node_modules are installed, or set the root of Nx Console to point to a Nx workspace in editor settings."
```

iguissouma: Maybe show error notification with two possible actions Run `pakageName` install, or show Nx settings to fix path ...

MaxKless: Yes we will look into setting up better error handling in the future. For now I just want to avoid a nondescriptive 'index out of bounds' error

Cammisuli (approved):

PR #1576, opened 2023-02-28
fix(intellij): move generate ui context menu to the new group

Cammisuli:


MaxKless (approved):
I think there could be an argument to remove the option from `ProjectViewPopupMenuModifyGroup` since it's in the context menu twice now. 

But since generators can do things other than new things it's actually fine. And since it's hidden in the `New` group, it won't clutter things up.

PR #1577, opened 2023-02-28
fix(nxls): do not include workspace layout in directory context for generate calls

Cammisuli:
When generating with the context menu, (depending where the context menu is initialized) sometimes we get the workspace layout directory prefilled. This causes generates to create a duplicated nested folders. (ie, `libs/libs`)

### DIFF
--- a/libs/language-server/workspace/src/lib/get-generator-context-from-path.ts
+++ b/libs/language-server/workspace/src/lib/get-generator-context-from-path.ts
@@ -34,11 +34,18 @@ export async function getGeneratorContextFromPath(
   const libsDir = workspaceLayout.libsDir;
   if (
     appsDir &&
-    (generator?.name === 'application' || generator?.name === 'app')
+    (generator?.name === 'application' ||
+      generator?.name === 'app' ||
+      modifiedPath.startsWith(appsDir))
   ) {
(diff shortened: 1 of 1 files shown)


MaxKless (approved):
I couldn't find the right generator to reproduce this with but it makes perfect sense as a fix. Only worry I have, maybe we should make the check include a trailing `/`. because if for example we had another folder path of `apps-backend` or something, right now it would replace it with `-backend`. 

With nested projects, crazier workspace setups are becoming more and more normal I think.

Cammisuli:
> I couldn't find the right generator to reproduce this with but it makes perfect sense as a fix. Only worry I have, maybe we should make the check include a trailing `/`. because if for example we had another folder path of `apps-backend` or something, right now it would replace it with `-backend`. 
> 
> 
> 
> With nested projects, crazier workspace setups are becoming more and more normal I think.

Then the whole path will be `appsDir/apps-backend`, and we'll still only remove the `appsDir` portion. 🙂 

If appsDir or libsDir is set then generators will most likely always place generated files in those directories. 

Nested projects shouldnt have appsDir or libsDir set.

We could try to include the `/`, but I'm pretty sure that already gets handled already.

PR #1584, opened 2023-03-02
feat(intellij): add nx run line marker contributor for project.json

iguissouma:
This PR will add gutter icon to project.json files to run nx targets on projects

Screenshot 2023-03-02 at 20 30 28

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/run/NxRunConfigurationProducer.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/run/NxRunConfigurationProducer.kt
@@ -1,0 +1,145 @@
+package dev.nx.console.run
+
+import com.intellij.execution.actions.ConfigurationContext
+import com.intellij.execution.actions.LazyRunConfigurationProducer
+import com.intellij.execution.configurations.ConfigurationFactory
+import com.intellij.json.psi.JsonFile
+import com.intellij.json.psi.JsonObject
+import com.intellij.json.psi.JsonProperty
+import com.intellij.json.psi.JsonStringLiteral
+import com.intellij.openapi.util.Ref
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.util.parentOfType
+
+data class NxRunSettings(
+    val nxProjects: String = "",
+    val nxTargets: String = "",
+    val arguments: String = ""
+)
+
+class NxRunConfigurationProducer : LazyRunConfigurationProducer<NxCommandConfiguration>() {
+    override fun getConfigurationFactory(): ConfigurationFactory =
+        NxCommandConfigurationType.getInstance()
+
+    override fun setupConfigurationFromContext(
+        configuration: NxCommandConfiguration,
+        context: ConfigurationContext,
+        sourceElement: Ref<PsiElement>
+    ): Boolean {
+
+        val runSettings =
+            createRunSettingsFromContext(
+                NxRunSettings(
+                    configuration.nxProjects,
+                    configuration.nxTargets,
+                    configuration.arguments
+                ),
+                context,
+                sourceElement
+            )
+
+        if (runSettings == null) {
+            return false
+        } else {
+            setupConfigurationFromSettings(configuration, runSettings)
+            return true
+        }
+    }
+
+    private fun setupConfigurationFromSettings(
+        configuration: NxCommandConfiguration,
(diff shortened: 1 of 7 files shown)


apps/intellij/src/main/kotlin/dev/nx/console/run/NxRunConfigurationProducer.kt
```diff
@@ -0,0 +1,145 @@
...
+        configuration.nxProjects = runSettings.nxProjects
+        configuration.nxTargets = runSettings.nxTargets
+        configuration.arguments = runSettings.arguments
+
+        // configuration.name = "${runSettings.nxProjects}[${runSettings.nxTargets}]"
+        configuration.setGeneratedName()
+    }
+
+    private fun createRunSettingsFromContext(
+        runSettings: NxRunSettings,
```
MaxKless, line 64: What's up with these two unused arguments here?

apps/intellij/src/main/kotlin/dev/nx/console/run/NxRunConfigurationProducer.kt
```diff
@@ -0,0 +1,145 @@
...
+        context: ConfigurationContext,
+        sourceElement: Ref<PsiElement>?
+    ): NxRunSettings? {
+        val element = getElement(context)
+        if (element != null && element.isValid) {
+            val psiProjectJsonFile = getContainingProjectJsonFile(element)
```
MaxKless, line 70: Do you think there's a way you could refactor this code a bit? Maybe by having some early return conditions instead of nesting conditionals deeper and deeper? I'm having some trouble reading the 4th nested `else` and I think folks in the future will too.

apps/intellij/src/main/kotlin/dev/nx/console/run/NxRunConfigurationProducer.kt
```diff
@@ -0,0 +1,145 @@
...
+                    return null
+                } else {
+                    val targetProperty = findContainingProperty(element)
+                    return if (targetProperty == null) {
+                        null
+                    } else {
+                        val propertyLiteral = element.parent as? JsonStringLiteral
+                        val nxTarget = propertyLiteral?.value ?: return null
+                        val nxProject =
+                            (((propertyLiteral.parent as? JsonProperty)
```
MaxKless, line 85: While this works, I think we should use the LSP as much as possible to parse JSON for us. Instead of walking up the tree here, you can use the (poorly named) `NxlsService.generatorContextFromPath` method to get information about the current project based on the path of the `project.json` file. 

We should rename that request to be more generic than "generator context" because really all it does is give you the project name and relative path based on a file path. 

@Cammisuli feel free to weigh in on quickly parsing manually vs using the nxls here.

MaxKless (changes requested):

apps/intellij/src/main/kotlin/dev/nx/console/run/NxRunConfigurationProducer.kt
iguissouma: I mimic the behaviour of NpmRunConfigurationProducer, the code is similar but sure it can be reworked

apps/intellij/src/main/kotlin/dev/nx/console/run/NxRunLineMarkerContributor.kt
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+fun isTargetProperty(property: JsonProperty?): Boolean {
+    val targetsProperty = PsiTreeUtil.getParentOfType(property, JsonProperty::class.java, true)
+    return targetsProperty != null && "targets" == targetsProperty.name
+}
+
+fun findContainingProperty(element: PsiElement): JsonProperty? {
+    return if (isInsideNxProjectJsonFile(element))
```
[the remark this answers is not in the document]
iguissouma: I reworked the code

g-cheishvili:
Does this PR also add a "debug" option to the targets or that's a separate feature planned sometime in the future?

iguissouma:
> Does this PR also add a "debug" option to the targets or that's a separate feature planned sometime in the future?

No debug support for now.

What do you want debug, the executor itself ?

MaxKless (approved):

MaxKless:
Hey @iguissouma i'm ready to merge this. can you rebase your changes and fix the conflict?
I did it locally but since it's your fork, I can't force push to the branch.

PR #1589, opened 2023-03-03
fix(intellij): refresh file system after running generator

### LINKED ISSUE #1586: [Intellij] force IDE to actualize Project Tree after a generator run
## Description

Currently, if I run any generator - IDE doesn't know about it so I don't see any updates in a Project Tree.

https://user-images.githubusercontent.com/1770529/222541621-ce0772d9-d7ce-46af-836f-e003cc4afba4.mp4

## Motivation

It will be a pretty common issue for JetBrains users.

## Suggested Implementation

Nx Console might call some JetBrains API method to trigger a `Reload from Disk` action (hope it exists).

## Alternate Implementations

I use a `Reload from Disk` action from the context menu to workaround it

MaxKless:
fixes #1586


apps/intellij/src/main/kotlin/dev/nx/console/generate/run_generator/RunGeneratorManager.kt
```diff
@@ -105,14 +107,33 @@ class RunGeneratorManager(val project: Project) {
...
     }
 }
+
+private fun List<String>.isDryRun(): Boolean {
+    return this.contains("--dry-run") || this.contains("--dryRun")
+}
+
+private fun String.isDryRun(): Boolean {
+    return this.contains("--dry-run") || this.contains("--dryRun")
+}
```
Cammisuli, line 139: I'm a little iffy on this. But we'll see in the future

Cammisuli (approved):
Merge the conflict, and it should be good to go 🙂

apps/intellij/src/main/kotlin/dev/nx/console/generate/run_generator/RunGeneratorManager.kt
MaxKless: you mean it shouldn't be an extension function?

PR #1596, opened 2023-03-08
chore(generate-ui): set font size via styling message

MaxKless:
here's some screenshots showing the new font sizes
image
image

image
image


Cammisuli (approved):
Looks good, just need to fix the formatting.

PR #1601, opened 2023-03-08
feat(intellij): add nx project toolwindow

iguissouma:
this PR will add a primary version of Nx Console toolWindow for running Nx Tasks

Screenshot 2023-03-08 at 22 47 59

Screenshot 2023-03-08 at 22 48 09


apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxExecutor.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.intellij.execution.RunnerAndConfigurationSettings
+import com.intellij.execution.executors.DefaultRunExecutor
+import com.intellij.execution.runners.ExecutionUtil
+import com.intellij.openapi.components.service
+import com.intellij.openapi.project.Project
+import dev.nx.console.run.NxCommandConfiguration
+import dev.nx.console.run.NxCommandConfigurationType
+import dev.nx.console.run.NxRunSettings
+
+class NxExecutor(val project: Project) {
```
MaxKless, line 14: In general it's good to have a class like this, but:
- it should live under `dev.nx.console.run` because we will reuse it for other things (e.g. the run target actions)
- it should be named differently because `Executor` is already a name in Nx-context. Maybe `NxTaskExecutionManager`?
- it should have a different signature. Since you only use one project and one target, I would do it like `fun execute(nxProject: NxProject, nxTarget: NxTarget)`

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
```diff
@@ -0,0 +1,327 @@
...
+
+            init {
+                icon = NxIcons.Action
+            }
+
+            override fun buildChildren(): Array<SimpleNode> {
+                if (nxWorkspace == null) {
+                    return emptyArray()
+                }
+                val tasks: Array<SimpleNode> = arrayOf(Tasks(nxWorkspace, this))
```
MaxKless, line 65: The Tree itself looks good. But we are moving more and more away from having these forced `apps` and `libs` directories. So it should be structured differently. The top-level nodes should be `Targets` and `Projects`.

The targets are already displayed well. For the projects, we have two options for displaying them: by folder structure or just having a list. In VSCode we call this "list view" and "tree view". 
You don't have to implement it like this right away, just pick one (list view is easier) and build it. Examples below 👇 

### Example of the tree view:
image
you can see the structure is the same as the folders

### Example of the list view:
image
you can see that everything is just one long list.

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
```diff
@@ -0,0 +1,327 @@
...
+                val apps: Array<SimpleNode> = arrayOf(Apps(nxWorkspace, this))
+                val libs: Array<SimpleNode> = arrayOf(Libs(nxWorkspace, this))
+                return tasks + apps + libs
+            }
+
+            // TODO do not add a root node or get the name of workspace properly?
```
MaxKless, line 71: I think adding a root node is a good idea, as it is the same as gradle toolwindow does it. It's an interesting question tho, how to get the name of a workspace. I think for now this is an okay solution. You could also use `project.nxBasepath` but it doesn't really matter.

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
```diff
@@ -0,0 +1,327 @@
...
+                nxWorkspace.workspace.projects.values
+                    .flatMap { p -> p.targets.keys.map { it to p.name } }
+                    .groupBy { it.first }
+                    .map { Task(nxWorkspace, it.key, this) }
+                    .toTypedArray()
+
+            override fun getName(): String = "Tasks"
+        }
+
+        class Task(private val nxWorkspace: NxWorkspace, val taskName: String, parent: SimpleNode) :
```
MaxKless, line 122: I would name these nodes 'targets' instead of tasks. They are not the same thing - a target is the definition in your `project.json` and is used to create a task which is a single invocation of a target. Here is a list of the targets defined in all the projects so it should be called targets.

In the future we will need nodes for configurations too but it doesn't have to be in this PR.

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
+
+    private inner class RunAction : ExecuteAction(DefaultRunExecutor.getRunExecutorInstance()) {
+        init {
+            shortcutSet = CommonShortcuts.ENTER
+        }
+    }
+
+    private inner class EditRunSettingsAction : AnAction(AllIcons.Actions.EditSource), DumbAware {
```
MaxKless, line 189: What do you think about splitting some of these classes out into separate files? I think having a 300 line file is not very readable and it would be easier to explore if for example the action code was separate.

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxToolWindowFactory.kt
```diff
@@ -0,0 +1,26 @@
...
+    override fun createToolWindowContent(project: Project, toolWindow: ToolWindow) {
+        val toolwindowPanel = NxToolWindowPanel(project)
+        val contentManager = toolWindow.contentManager
+        val content = contentManager.factory.createContent(toolwindowPanel, null, false)
+        content.isCloseable = false
+        contentManager.addContent(content)
+    }
+}
+
+fun Project.nxWorkspace() = runBlocking { service<NxlsService>().workspace() }
```
MaxKless, line 20: I don't know about having an extension function here. For `nxBasePath` is makes sense because we always use it instead of `project.basePath`. But here we only hide the fact that we're calling the `NxlsService`. Imo it would be better to have it explicit but feel free to explain why you did it this way. @Cammisuli you can also weigh in.

apps/intellij/src/main/resources/META-INF/plugin.xml
```diff
@@ -59,6 +59,15 @@
...
 
+    <!--nx workspace service-->
+    <projectService serviceImplementation="dev.nx.console.services.NxWorkspaceService"/>
+
+
+    <!--Tool Window-->
+    <toolWindow id="Nx Console"
+                anchor="right"
+                factoryClass="dev.nx.console.toolWindow.NxToolWindowFactory"
+                icon="/icons/nx-console.svg" />
```
MaxKless, line 70: Does this have to an `.svg` or can we use the `NxIcons.Action` class here too?

MaxKless (changes requested):

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxToolWindowFactory.kt
iguissouma: just found it an easy way to get the nxWorkspace without much code we can refactor it and inline it...

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
iguissouma: They are inner classes they access to fields of the outer class and they are not intended to be used outside...
I saw this pattern while navigating on intellij code base

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
iguissouma: I was using the value npmScope from the json object but not sure if it's the intended value
We can use maybe project.name when nx monorepo is the root...
But let keep it simple

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
iguissouma: ok for naming targets

Yeah I tried the run configuration but doesn't work well...

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
iguissouma: They are inner classes, they access outer class fields like the tree to reload it for example...

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
iguissouma: the apps and lib are not the directories but rather the types of nx projects
we can have some settings that says for Projects group by project type, by path ...

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxToolWindowFactory.kt
MaxKless: yeah it is pretty short that's true

apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxProjectsTreeStructure.kt
MaxKless: yeah that makes sense... let's leave it like this for now

PR #1609, opened 2023-03-14
feat(intellij): add nx workspace service to manage workspace changes

iguissouma:
This PR will add a first support to manage nx workspace state


MaxKless (changes requested):
Good solution but two questions:
- why did you make it into a separate `NxWorkspaceService`? I think it could just live in `NxlsService` so it's all in one place.
- shouldn't a message be published to the message bus whenever a refresh notification comes from the nxls? I would add a call here: `NxlsLanguageClient#refreshWorkspace` so a message is sent every time a file changes.

PR #1611, opened 2023-03-15
fix(intellij): don't apply empty workspace path setting

MaxKless:


Cammisuli (approved):

apps/intellij/src/main/kotlin/dev/nx/console/nxls/WorkspacePathSetting.kt
```diff
@@ -42,6 +42,10 @@ class WorkspacePathSetting(val project: Project) : NxConsoleSettingBase<String?>
     }
 
     override fun getValue(): String? {
+        val value = this.inputField.text
```
iguissouma, line 45: It can be one line 
`override fun getValue(): String? = inputField.text.ifEmpty { null }`

MaxKless: yeah that looks better

PR #1613, opened 2023-03-15
feat(intellij): add nx console project toolwindow

iguissouma:


apps/intellij/src/main/kotlin/dev/nx/console/toolWindow/NxExecutor.kt
```diff
@@ -0,0 +1,46 @@
...
+                    val nxCommandConfiguration = it.configuration as NxCommandConfiguration
+                    val nxRunSettings = nxCommandConfiguration.nxRunSettings
+                    nxRunSettings.nxTargets == nxTarget && nxRunSettings.nxProjects == nxProject
+                }
+                ?: runManager
+                    .createConfiguration(
+                        "$nxProject[$nxTarget]",
+                        NxCommandConfigurationType::class.java
+                    )
+                    .apply {
```
bulldog98, line 32: would the cast not be better outside of the apply? Also two apply inside each other are not good in the readability sense

MaxKless: hmm you're not wrong

MaxKless (approved):

Cammisuli (approved):
This is awesome! Thank you again @iguissouma

PR #1615, opened 2023-03-15
fix(vscode): handle daemon better

Cammisuli:
* Adds a flag to only send notifications to the language server if the language server didn't send the notification first
* Removes the cache for finding local nx packages
* Set `NX_WORKSPACE_ROOT_PATH`


MaxKless (approved):

MaxKless:
If you

Cammisuli:
Good call, i changed it to nxls 🙂

PR #1623, opened 2023-03-18
feat(intellij): add nx generator search evrywhere contributor

iguissouma:
this PR will add nx generator search evrywhere contributor

Screenshot 2023-03-18 at 16 01 30

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
@@ -1,0 +1,121 @@
+package dev.nx.console.generate
+
+import com.intellij.icons.AllIcons
+import com.intellij.ide.actions.searcheverywhere.SearchEverywhereContributor
+import com.intellij.ide.actions.searcheverywhere.SearchEverywhereContributorFactory
+import com.intellij.openapi.actionSystem.ActionPlaces
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.CommonDataKeys
+import com.intellij.openapi.components.service
+import com.intellij.openapi.fileEditor.FileEditorManager
+import com.intellij.openapi.progress.ProgressIndicator
+import com.intellij.openapi.progress.ProgressManager
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.psi.codeStyle.NameUtil
+import com.intellij.util.Processor
+import dev.nx.console.generate.ui.DefaultNxGenerateUiFile
+import dev.nx.console.generate.ui.NxGeneratorListCellRenderer
+import dev.nx.console.models.NxGenerator
+import dev.nx.console.models.NxGeneratorOption
+import dev.nx.console.nxls.server.requests.NxGeneratorOptionsRequestOptions
+import dev.nx.console.services.NxlsService
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+import javax.swing.ListCellRenderer
+import kotlinx.coroutines.runBlocking
+
+class NxGeneratorSearchEvrywhereContributor : SearchEverywhereContributorFactory<NxGenerator> {
+    override fun createContributor(
+        initEvent: AnActionEvent
+    ): SearchEverywhereContributor<NxGenerator> = NxGeneratorSEContributor(initEvent)
+}
+
+class NxGeneratorSEContributor(val event: AnActionEvent) :
+    SearchEverywhereContributor<NxGenerator> {
+
+    private val project = event.getRequiredData(CommonDataKeys.PROJECT)
+
+    private val generators = runBlocking { project.service<NxlsService>().generators() }
+
+    private val generatorToOptionsCache: ConcurrentMap<NxGenerator, List<NxGeneratorOption>> =
+        ConcurrentHashMap()
+
+    override fun getSearchProviderId(): String = javaClass.name
+
+    override fun getGroupName(): String = "Nx"
+
+    override fun getSortWeight(): Int = 0
+
+    override fun showInFindResults(): Boolean = true
+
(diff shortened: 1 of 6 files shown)


apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
```diff
@@ -0,0 +1,121 @@
...
+import dev.nx.console.models.NxGenerator
+import dev.nx.console.models.NxGeneratorOption
+import dev.nx.console.nxls.server.requests.NxGeneratorOptionsRequestOptions
+import dev.nx.console.services.NxlsService
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+import javax.swing.ListCellRenderer
+import kotlinx.coroutines.runBlocking
+
+class NxGeneratorSearchEvrywhereContributor : SearchEverywhereContributorFactory<NxGenerator> {
```
MaxKless, line 27: Shouldn't this be `NxGeneratorSearchEverywhereContributorFactory`? Also there is an e missing in everywhere

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
```diff
@@ -0,0 +1,121 @@
...
+    override fun createContributor(
+        initEvent: AnActionEvent
+    ): SearchEverywhereContributor<NxGenerator> = NxGeneratorSEContributor(initEvent)
+}
+
+class NxGeneratorSEContributor(val event: AnActionEvent) :
```
MaxKless, line 33: This can be spelled out if the factory class is renamed.

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
```diff
@@ -0,0 +1,121 @@
...
+                null
+            }
+
+        val generatorWithOptions = NxGenerator(selected, generatorOptions)
+
+        val generatorContext = runBlocking {
+            path?.let { service.generatorContextFromPath(generatorWithOptions, path) }
+        }
+
+        val virtualFile = DefaultNxGenerateUiFile("Generate", project)
```
MaxKless, line 89: Since this is the same code as in the `Nx Generate (UI)` action, we could reuse it by moving it to the `NxGenerateService`

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    override fun getGroupName(): String = "Nx"
+
+    override fun getSortWeight(): Int = 0
+
+    override fun showInFindResults(): Boolean = true
+
+    override fun getElementsRenderer(): ListCellRenderer<in NxGenerator> =
+        NxGeneratorListCellRenderer(
+            myIcon = AllIcons.Actions.GeneratedFolder,
```
MaxKless, line 53: I like the `new folder` icon, but I'm not sure if it wouldn't be more consistent to use the nx console icon everywhere for visual clarity. Just a thought, idk what you think @Cammisuli @iguissouma

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        if (StringUtil.isEmptyOrSpaces(pattern)) {
+            return
+        }
+
+        val matcher = NameUtil.buildMatcher(pattern).build()
+
+        ProgressManager.getInstance()
+            .executeProcessUnderProgress(
+                { generators.filter { matcher.matches(it.name) }.forEach { consumer.process(it) } },
```
MaxKless, line 117: how about allowing input like `nx g @nrwl/xxx` or something? so it's more like what you would type in the CLI?

MaxKless (changes requested):

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
iguissouma: I'm aware of this duplicate code, I will see if I can refactor it.

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
iguissouma: I like the idea of a new icon that is different from the Action, the Action icon will be used for main Actions, not repeated for all generators...

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
iguissouma: We should allow searching for generator like any symbols(a class, a table), and for this item I can for example execute an action, Open in UI or On Run Anything... 
That was the idea

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
MaxKless: should we add that too to the run anything generators?

apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGeneratorSearchEvrywhereContributor.kt
iguissouma: I reused the openGenerateUi method for the moment

PR #1625, opened 2023-03-20
feat(vscode): move analytics to measurement protocol

Cammisuli:
- feat(vscode): use measurement protocol for telemetry
- remove user class and use env.isTelemetryEnabled


MaxKless (approved):
Looks good functionality-wise. 
We should probably add that list of telemetry identifiers we talked about & make everything consistent. e.g. the cloud webview uses telemetry too, not sure how we should structure those identifiers.

PR #1644, opened 2023-03-30
fix(nxls): do not output exec to stdio (#1644

Cammisuli:
## What it does
In vscode, nxls is started via a fork and communicates with ipc. So messages on stdio will not affect the communication.
With Intellij, the nxls is started with a separate process and communicates with stdio. This means that any `exec` calls that are called within the language server could pollute the stdio communicate and cause the process to exit. So we have to make sure that we set exec calls to `stdio: ignore`.

This also fixes the nx graph to show up properly in Windows.

### DIFF
--- a/apps/intellij/build.gradle.kts
+++ b/apps/intellij/build.gradle.kts
@@ -173,12 +173,16 @@ tasks {
 }
 
 tasks.register<Exec>("buildNxls") {
-    commandLine = listOf("bash", "-c", "npx nx run nxls:build")
+    commandLine = buildCommands().apply {
+        add("npx nx run nxls:build")
+    }
     workingDir = rootDir
 }
 
 tasks.register<Exec>("buildGenerateUi") {
-    commandLine = listOf("bash", "-c", "npx nx run generate-ui:build:production-intellij")
+    commandLine = buildCommands().apply {
+        add("npx nx run generate-ui:build:production-intellij")
+    }
     workingDir = rootDir
 }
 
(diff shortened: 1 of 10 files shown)


apps/intellij/build.gradle.kts
```diff
@@ -194,3 +198,11 @@ tasks.register<DefaultTask>("publish") {
     group = "publish"
     description = "Placeholder task to workaround the semantic-release plugin"
 }
+
+fun buildCommands() = mutableListOf<String>().apply {
```
iguissouma, line 202: I think there's no need of creating mutableListOf<String>() and use function apply

`fun buildCommands() = if(isOsWindows()) listOf("pwsh", "-command") else listOf("bash", "-c")`

if list should be mutable yo can replace listOf to mutableListOf

apps/intellij/build.gradle.kts
```diff
@@ -173,12 +173,16 @@ tasks {
...
-    commandLine = listOf("bash", "-c", "npx nx run nxls:build")
+    commandLine = buildCommands().apply {
+        add("npx nx run nxls:build")
+    }
     workingDir = rootDir
 }
 
 tasks.register<Exec>("buildGenerateUi") {
-    commandLine = listOf("bash", "-c", "npx nx run generate-ui:build:production-intellij")
+    commandLine = buildCommands().apply {
```
iguissouma, line 183: Maybe this can be the same as
`buildCommands().plus("npx nx run generate-ui:build:production-intellij")`

apps/intellij/build.gradle.kts
Cammisuli: Now this is just me being fairly new to Kotlin, but why do it this way instead of using `apply`? I'm just curious 😊

iguissouma: Nothing wrong, more concise code :)

PR #1686, opened 2023-04-18
feat: support nx 16 

Cammisuli:
Also updates nx-console to v16


libs/language-server/workspace/src/lib/read-collections.ts
```diff
@@ -208,7 +208,7 @@ export async function getCollectionInfo(
     const extendedCollections = (
       await Promise.all(
         extendedSchema
-          .filter((extended) => extended !== '@nrwl/workspace')
+          .filter((extended) => extended !== '@nx/workspace')
```
MaxKless, line 211: won't this line be executed in both v16 and earlier versions? so shouldn't we handle both cases here?

libs/shared/npm/src/lib/find-nx-package-path.ts
```diff
@@ -25,7 +25,7 @@ export async function findNxPackagePath(
 
   const nrwlWorkspaceDepPath = await workspaceDependencyPath(
     workspacePath,
-    '@nrwl/workspace'
+    '@nx/workspace'
```
MaxKless, line 28: here we might also need another case for those not yet on v16 but also without an `nx` package? not sure..

libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
```diff
@@ -233,16 +233,14 @@ export class NxCloudService extends StateBaseService<InternalState> {
           try {
             if (lt(coerce(nxVersion.full) ?? '', '12.0.0')) {
               promisify(exec)(
-                `${getPackageManagerCommand().addDev} @nrwl/nx-cloud@${
+                `${getPackageManagerCommand().addDev} @nx/nx-cloud@${
```
MaxKless, line 236: this is not going to work because there's not `@nx/nx-cloud` with v12 right?

libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
```diff
@@ -468,7 +466,7 @@ export class NxCloudService extends StateBaseService<InternalState> {
   private async loadAndSetIsPrivateCloud() {
     const nxWorkspaceConfig = (await getNxWorkspace()).workspace;
     const cloudRunnerUrl =
-      nxWorkspaceConfig.tasksRunnerOptions?.default?.runner === '@nrwl/nx-cloud'
```
MaxKless, line 471: here we need to check for both `@nrwl/nxcloud` and `nx-cloud` runners because this feature should work regardless of your nx version.

libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
```diff
@@ -668,7 +666,7 @@ export class NxCloudService extends StateBaseService<InternalState> {
       return false;
     }
     return !!Object.values(nxConfig.tasksRunnerOptions).find(
-      (r) => r.runner == '@nrwl/nx-cloud'
+      (r) => r.runner == '@nx/nx-cloud'
```
MaxKless, line 669: same here

libs/vscode/nx-workspace/src/lib/get-nx-workspace.ts
```diff
@@ -31,7 +31,7 @@ export async function getNxCloudRunnerOptions(): Promise<
   }
   const nxCloudTaskRunner = Object.values(
     workspaceConfig.tasksRunnerOptions
-  ).find((r) => r.runner == '@nrwl/nx-cloud');
+  ).find((r) => r.runner == '@nx/nx-cloud');
```
MaxKless, line 34: also need to handle both cases

MaxKless (changes requested):

libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
Cammisuli: you're right. The migration that changed @nrwl to @nx was a bit overzealous, and I'm going to try the migration again with the latest beta today.

MaxKless (approved):

PR #1712, opened 2023-04-28
chore(vscode): revert faulty external change in cloud view project.json

MaxKless:


Cammisuli (approved):
just reformat it and it should be good to go

PR #1716, opened 2023-05-02
feat(intellij): add generator filter setting

### LINKED ISSUE #1703: Limit list of generators and executors
## Description

Configuration Options to limit the generators and executors offered for execution to the user by the interface.

## Motivation

In our Workflow we use a lot of custom generators that built on top of default generators like @nrwl/nest to provide structuring and additional functionality to the project, like built in features for Docker. These default generators should therefore never be used by our developers. However they can not be made invisible at the moment. This results in unnecessary confusion. 

In addition to that we use executors that are only intended for use inside of the CI-System, like for example a step to publish Containers or apply Infrastructure as Code. Their appearance in the interface is not needed as well.

The implementation of these features would reduce the complexity of the developer experience, especially with larger teams.

## Suggested Implementation

A possible way to limit the generators would be to modify the language Server to limit the given list of generators to a list of whitelisted plugins in the nx.json. After adding an enabled : false attribute to the plugins not needed in the generators section, the generators should be removed from the list. This format is compatible with the existing schema.

### Example
```
{
  "$schema": ...,
  "npmScope": ..,
  "plugins": [],
  "tasksRunnerOptions": {...}
  "generators": {
    "@nrwl/react": {
      enabled": false,
      "application": {
        "babel": true
      },
      "library": {
        "unitTestRunner": "jest"
      }
    },
  }
}
```
As for available executors the attribute CI could be added to the executor in the project.json file. If this is set to true, the executor should not be displayed in the interface.

### Example

```
"publish": {
      "executor": "...",
      "ci": true,
      "options": {
        "outDir": "..."
      }
}
```
Thanks for your Feedback :)

MaxKless:
fixes #1703 

image
image

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/settings/NxConsoleProjectSettingsProvider.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/settings/NxConsoleProjectSettingsProvider.kt
@@ -4,7 +4,7 @@ import com.intellij.openapi.components.*
 import com.intellij.openapi.project.Project
 import dev.nx.console.settings.options.GeneratorFilter
 
-@State(name = "NxConsoleSettingsProvider", storages = [Storage("nx-console.xml")])
+@State(name = "NxConsoleProjectSettingsProvider", storages = [Storage("nx-console.xml")])
 class NxConsoleProjectSettingsProvider(val project: Project) :
     PersistentStateComponent<NxConsoleProjectSettingsState> {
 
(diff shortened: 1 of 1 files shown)


apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGenerateService.kt
```diff
@@ -118,3 +152,13 @@ class NxGenerateService(val project: Project) {
             project.getService(NxGenerateService::class.java)
     }
 }
+
+private fun matchWithWildcards(text: String, pattern: String): Boolean {
+    val regexFromPattern =
```
Cammisuli, line 157: I wonder if we can use some kind of matching library rather than do this regex ourselves.

MaxKless: I looked around for a bit and didn't find anything that looked good - not sure if it's worth the effort.

Cammisuli: yea, good point

Cammisuli (approved):

PR #1744, opened 2023-05-22
chore: rename build-webview-files target to build

MaxKless:
this was a relic from the dev process, there's no reason it should have this special name anymore


MaxKless:
nvm there was a reason 😆

PR #1749, opened 2023-05-26
feat(generate-ui): rewrite generate ui and display it via feature toggle

MaxKless:


libs/vscode/configuration/src/lib/configuration-keys.ts
```diff
@@ -9,6 +9,7 @@ export const GLOBAL_CONFIG_KEYS = [
   'enableTaskExecutionDryRunOnChange',
   'projectViewingStyle',
   'moveGeneratorPatterns',
+  'useNewGenerateUIPreview',
```
Cammisuli, line 12: Does the case matter here now that the package.json is using `useNewGenerateUiPreview`?

Cammisuli (approved):
LGTM - just one concern about the name of the setting. And also fix the formatting 😄

PR #1752, opened 2023-06-01
feat: support nx 16.3

Cammisuli:


migrations.json
MaxKless, line 1: This should probably be removed from the repo right?

MaxKless (changes requested):
Tested it with a new `16.3.2` workspace and it threw an error right away: 
```
Unable to get nx workspace configuration: TypeError: nxProjectGraphUtils?.createProjectFileMapUsingProjectGraph is not a function

MaxKless (approved):

PR #1777, opened 2023-06-21
feat: refactor nx cli commands, throw out deprecated or outdated stuff

MaxKless:
reworks how the cli quickpick commands are handled (`nx.generate`, `nx.run`, `nx.run.target`, `nx.affected` etc).
There's now a separate lib for all the quickpick related code. 
Before it was mixed into the tasks lib, which resulted in de-facto circular dependencies (`vscode-tasks` has commands that call generate ui, generate ui runs tasks in `vscode-tasks`). Those never surfaced because it's all abstracted away behind vscode commands. Now it's resolved because the parts that both need are in `vscode-nx-cli-quickpicks`

The `nx.run` and `nx.run.target` actions do the same thing but in different orders. Still, they had divergent implementations. This is now consolidated into one function. A bunch of legacy code that dealt with a different scope (i.e. that function was still used to generate stuff too) has been removed at the same time. 

Some deprecated/rarely used actions have been removed: `nx.affected.apps`, `nx.affected.libs` (because the nx functionality doesn't exist anymore), `nx.generate.apps/libs` and their variants because barely anyone used them, they provide very little value and are overhead to maintain. 

Other small fixes have been made while I went.


libs/vscode/nx-cli-quickpicks/src/lib/select-affected-flags.ts
```diff
@@ -0,0 +1,106 @@
+import { OptionType, Option } from '@nx-console/shared/schema';
+import { selectFlags } from './select-flags';
+
+/**
+ *
+ */
```
Cammisuli, line 6: Was something important missing here?

libs/vscode/tasks/tsconfig.lib.json
```diff
@@ -7,5 +7,12 @@
...
   },
   "exclude": ["**/*.spec.ts", "**/*.test.ts", "jest.config.ts"],
-  "include": ["**/*.ts"]
+  "include": [
+    "**/*.ts",
+    "../nx-cli-quickpicks/src/lib/select-generator.ts",
+    "../nx-cli-quickpicks/src/lib/select-re-move-generator.ts",
+    "../nx-cli-quickpicks/src/lib/cli-task-quick-pick-item.ts",
+    "../nx-cli-quickpicks/src/lib/select-flags.ts",
+    "../nx-cli-quickpicks/src/lib/cli-task-flag-quick-pick-item.ts"
```
Cammisuli, line 16: remove these

package.json
```diff
@@ -65,7 +65,7 @@
     "@nx/esbuild": "16.3.1",
     "@nx/eslint-plugin": "16.3.1",
     "@nx/jest": "16.3.1",
-    "@nx/js": "16.3.1",
+    "@nx/js": "16.3.2",
```
Cammisuli, line 68: Why was this updated but not the other packages?

libs/vscode/nx-cli-quickpicks/src/lib/select-affected-flags.ts
MaxKless: nah this empty comment block was there before I just copied it over.

package.json
MaxKless: curious, I'm pulling everything up to `16.4.0`. Tested across both IDEs and it seems there were no problems.

MaxKless:
I also threw out `generate-ui-e2e` since it was causing issues and we're removing it soon anyways. No need to spend time and energy on fixing it imo.

Cammisuli (approved):

PR #1795, opened 2023-06-29
fix: make sure nx.run works in all contexts

MaxKless:


Cammisuli (approved):

mechanoid7:
why are you changed  
let p = projectName;
let t = targetName; ?
what's the problem with prev ver?

MaxKless:
i think i didn't want to reassign the parameters, why?

PR #1796, opened 2023-06-29
fix: revert ts version because it broke old generate ui

MaxKless:


Cammisuli (approved):
LGTM, things don't seem to be broken :smile:

PR #1797, opened 2023-06-30
feat(vscode): rearrange empty state for nx cloud view to make button more prominent

MaxKless:
This is what it looks like now: 
image

Before, the button was hidden if all other views were open (the default state). Now you can see the button either way: 
image


libs/vscode/nx-cloud-view/src/webview/main.ts
```diff
@@ -69,6 +78,7 @@ export class Root extends LitElement {
   private setState(state: WebviewState) {
     this.state = { ...state };
     this.vscodeApi.setState(state);
+    console.log(this.state);
```
Cammisuli, line 81: Do you still want this here?

Cammisuli (approved):

PR #1823, opened 2023-07-14
feat: consume new graph events to open files in the ide

MaxKless:
This will enable the opening of files by clicking on project edge links in the integrated graph.
We set `window.environment = 'nx-console'` to let the graph know it's rendered in nx console. 
Because this wasn't the case before and would break some things, it's only enabled from 17.0.0.

Once the nx change is released, we can change this number to reflect the actual release.

Corresponding nx PR: https://github.com/nrwl/nx/pull/18113

Let's wait until it's reviewed before merging.

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/graph/ui/NxGraphBrowser.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/graph/ui/NxGraphBrowser.kt
@@ -155,8 +155,7 @@ class NxGraphBrowser(
 
         val nxConsoleEnvironmentScriptTag: String =
             NxlsService.getInstance(project).nxVersion()?.let {
-                // TODO: once this is released on the nx side, replace with the proper version check
-                if (it.full.toVersion(strict = false) >= "17.0.0".toVersion(strict = false)) {
+                if (it.full.toVersion(strict = false) >= "16.6.0".toVersion(strict = false)) {
                     "<script> window.environment = 'nx-console' </script>"
                 } else {
                     ""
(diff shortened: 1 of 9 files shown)


apps/intellij/src/main/kotlin/dev/nx/console/graph/ui/NxGraphBrowser.kt
```diff
@@ -142,6 +153,17 @@ class NxGraphBrowser(
 
         val basePath = "${Path(fullPath).parent}/"
 
+        val nxConsoleEnvironmentScriptTag: String =
+            NxlsService.getInstance(project).nxVersion()?.let {
+                // TODO: once this is released on the nx side, replace with the proper version check
+                if (it.full.toVersion(strict = false) >= "17.0.0".toVersion(strict = false)) {
```
Cammisuli, line 159: This change will be in `16.6.0`

libs/vscode/project-graph/src/lib/load-html.ts
```diff
@@ -373,3 +387,27 @@ function injectedScript() {
     }())
   `;
 }
+
+async function setNxConsoleEnvironment() {
+  const nxVersion = await getNxVersion();
+  console.log(nxVersion);
+  // TODO: once this is released on the nx side, replace with the proper version check
+  if (gte(nxVersion.version, '17.0.0')) {
+    console.log('hello');
```
Cammisuli, line 396: rm the console log when you change the version

Cammisuli (approved):
Change the version checks to `16.6.0` and remove the console.log, otherwise LGTM

PR #1825, opened 2023-07-18
fix(generate-ui): filter autocomplete options using .includes instead of .startsWith

MaxKless:


LayZeeDK:
Inline auto completion isn't a must-have.

rubenheymans:
with you first option you have to type to much, second option you just need to type the last part of a path and you have your result in the list

Cammisuli (approved):

PR #1860, opened 2023-08-17
feat: move plugins to lsp & write internal projectRootAndNameFormat plugin

MaxKless:
- move generator schema transformation to lsp
- add capacity for fixed form values
- fix fixed form value handling and add internal pnar plugin
- plugin to display hints and show them in generate ui
- apply schemaProcessor plugins in intellij


Cammisuli (approved):

MaxKless:
For now just use the command line @dshpindler. We're rethinking this PR though, maybe it will be added back to Nx Console soon. In the future,`projectNameAndRootFormat=derived` will be dropped, though. 
So sooner or later, `as-provided` will be the way to go.

dshpindler:
thank you @MaxKless! could you explain why is it dropped? and what is the alternative?

I only care about having the import path to be the same as the lib's path.
(e.g `@org/my-app/web/cluster-info/data-access` when derived vs `@org/data-access` when `as-provided`)

MaxKless:
Check out the Nx live stream next week, we'll be talking about the name / dir option updates and give some context!

PR #1866, opened 2023-08-24
fix(intellij): add shortcut to focus task action

MaxKless:
there's now a shortcut in the gutter and the toolwindow
image
image

`ctrl shift g` because it's the same as `ctrl shift r` for running a target but `g` for graph.

There's something funky going on breaking shortcuts when you open the gutter actions with `option enter`... but that seems like a separate concern from this PR.

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/graph/actions/NxGraphFocusTaskAction.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/graph/actions/NxGraphFocusTaskAction.kt
@@ -3,6 +3,7 @@ package dev.nx.console.graph.actions
 import com.intellij.openapi.actionSystem.*
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.project.DumbAwareAction
+import dev.nx.console.NxIcons
 import dev.nx.console.graph.NxGraphService
 import dev.nx.console.nx_toolwindow.tree.NxSimpleNode
 import dev.nx.console.nx_toolwindow.tree.NxTreeNodeKey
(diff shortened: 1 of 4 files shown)


apps/intellij/src/main/kotlin/dev/nx/console/graph/actions/NxGraphFocusTaskAction.kt
```diff
@@ -20,6 +18,10 @@ import kotlinx.coroutines.*
 class NxGraphFocusTaskAction(private val targetDescriptor: NxTargetDescriptor? = null) :
     DumbAwareAction() {
 
+    init {
+        shortcutSet = CustomShortcutSet.fromString("control shift G")
```
Cammisuli, line 22: We should find a way to make this customizable via the keyboard map settings

Cammisuli (changes requested):

apps/intellij/src/main/kotlin/dev/nx/console/graph/actions/NxGraphFocusTaskAction.kt
MaxKless: yeah fair enough. I can find some other predefined shortcut set. Any suggestions on the key combo?

Cammisuli: You can keep the same combo, but just allow it to be configured with a custom keyboard shortcut.

MaxKless: done

apps/intellij/src/main/kotlin/dev/nx/console/graph/ui/NxGraphBrowser.kt
```diff
@@ -457,6 +457,7 @@ class NxGraphBrowser(
             val js =
                 """
             window.externalApi?.registerFileClickCallback?.((message) => {
+            console.log("file click", message)
```
Cammisuli, line 460: Are these console.log's needed?

Cammisuli (approved):

PR #1869, opened 2023-08-28
feat: write internal git clean startup message plugin & display startup messages in intellij

MaxKless:
moves startup message plugin handling to nxls to enable it in both vscode & intellij. You can write plugins to display startup messages in 
Creates an internal plugin to dynamically display a warning message when generating with unstaged files.

image
image


libs/shared/nx-console-plugins/src/lib/internal-plugins/git-clean-message-factory.ts
```diff
@@ -0,0 +1,24 @@
...
+import { exec } from 'child_process';
+
+export async function gitCleanMessageFactory(
+  _: GeneratorSchema,
+  workspace: NxWorkspace
+): Promise<StartupMessageDefinition | undefined> {
+  const workspacePath = workspace.workspacePath;
+  try {
+    await promisify(exec)('git diff --quiet', {
+      cwd: workspacePath,
```
Cammisuli, line 14:
```suggestion
      cwd: workspacePath,
      windowsHide: true
```

Cammisuli (approved):

PR #1873, opened 2023-08-31
fix: move projectNameAndRootFormat message to banner from tooltip

MaxKless:
image


libs/shared/nx-console-plugins/src/lib/internal-plugins/project-name-and-root-plugin.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (
+    !schema?.options?.find(
+      (option) => option.name === 'projectNameAndRootFormat'
+    )
+  ) {
+    return undefined;
+  }
+  return {
+    message:
+      'Starting with Nx 16.7, Nx Console will generate projects with the exact name and directory provided. Check the dry-run output to see the files which will be created.',
```
Cammisuli, line 54:
```suggestion
      'Starting with Nx 16.7, Nx Console will generate projects with the exact name and directory provided. Check the output files to make sure that they were created in the correct location.',
```

Cammisuli (changes requested):
reword

Cammisuli (approved):

PR #1904, opened 2023-10-16
feat(generate-ui): add cwd handling & prefilling

MaxKless:
This PR sets up Nx Console to better deal with the new generator setup in Nx 17. 
- We can set a cwd value in the form and pass it along to VSCode & IntelliJ
- The cwd value will be prefilled instead of the directory
- The nameAndDirectoryFormat option will be prefilled to as-provided and a banner message for information will be shown
- hints on the name property let users know you can provide nested names

### DIFF
--- a/libs/shared/nx-console-plugins/src/lib/internal-plugins/name-and-directory.plugin.ts
+++ b/libs/shared/nx-console-plugins/src/lib/internal-plugins/name-and-directory.plugin.ts
@@ -22,6 +22,8 @@ export const nameAndDirectoryProcessor: SchemaProcessor = (
         return {
           ...option,
           'x-priority': 'important',
+          'x-hint':
+            'You can provide a nested name instead of setting the directory option, e.g. my-dir/my-component',
         };
       }
       if (option.name === 'directory') {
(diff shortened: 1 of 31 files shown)


libs/shared/nx-console-plugins/src/lib/internal-plugins/name-and-directory.plugin.ts
```diff
@@ -0,0 +1,71 @@
...
+) => {
+  if (
+    !schema?.options?.find((option) => option.name === 'nameAndDirectoryFormat')
+  ) {
+    return undefined;
+  }
+
+  return {
+    message:
+      'Starting with Nx 17, Nx Console will generate artifacts with the exact name and directory provided. Check the output files to make sure that they were created in the correct location. You can revert to the old behavior by updating nameAndDirectoryFormat',
```
Cammisuli, line 68: The last sentence should be a little more clearer. "You can revert to the old behaviour by updating nameAndDirectoryFormat". Update it where? Nx.json? project.json?

Cammisuli (commented):
There's also some weird behaviour that I saw. We could probably sync up and see if there's something weird I'm doing.

Cammisuli (approved):

PR #1908, opened 2023-10-17
fix(vscode): handle new nx cloud access token format in v17

MaxKless:
In nx17, nx cloud can be configured without the `taskRunnerOptions` object. Nx Console can now deal with this format.

### DIFF
--- a/libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
+++ b/libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
@@ -10,7 +10,7 @@ import {
   getWorkspacePath,
   watchFile,
 } from '@nx-console/vscode/utils';
-import { exec, spawn } from 'child_process';
+import { exec, execSync, spawn } from 'child_process';
 import {
   debounceTime,
   distinctUntilChanged,
@@ -264,7 +264,7 @@ export class NxCloudService extends StateBaseService<InternalState> {
                 'nx',
                 // https://github.com/nrwl/nx/pull/12942
                 ...(gte(coerce(nxVersion.full) ?? '', '15.0.7')
-                  ? ['connect', '--interactive', 'false']
+                  ? ['connect']
                   : ['connect-to-nx-cloud']),
               ],
               { cwd: getWorkspacePath(), env, shell: true }
--- a/libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
+++ b/libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
@@ -422,6 +422,20 @@ export class NxCloudService extends StateBaseService<InternalState> {
       );
       return;
     }
+    if (!this.state.canAccessCloudWorkspace) {
+      window
+        .showErrorMessage(
+          "You don't have access to this workspace. Log in to Nx Cloud in order to set up VCS integration.",
+          'Login',
+          'Cancel'
+        )
+        .then((value) => {
+          if (value === 'Login') {
+            commands.executeCommand('nxConsole.loginToNxCloud');
+          }
+        });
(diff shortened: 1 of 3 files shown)


libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
```diff
@@ -459,27 +460,22 @@ export class NxCloudService extends StateBaseService<InternalState> {
...
+
+    this.setState({ cloudRunnerUrl: cloudRunnerOptions?.url });
   }
 
   private async generateListOfFirstCommands(): Promise<void> {
     const generateAndSetFirstCommands = async () => {
       const nxConfig = (await getNxWorkspace()).workspace;
-      const cacheableOperations =
-        nxConfig.tasksRunnerOptions?.default?.options.cacheableOperations;
+      const cacheableOperations = await getCacheableOperations();
```
AgentEnder, line 477: What do we use this for? We are working towards `cache` being settable on a per target basis, so this may not be accurate soon.

libs/vscode/nx-workspace/src/lib/get-nx-workspace.ts
```diff
@@ -61,3 +77,28 @@ function getNxCloudEnv(workspacePath: string): any {
...
+    if (!workspace.projects.hasOwnProperty(key)) {
+      continue;
+    }
+    const targets = workspace.projects[key].targets;
+    for (const targetKey in targets) {
+      if (!targets?.hasOwnProperty(targetKey)) {
+        continue;
+      }
+      // TODO: remove type cast after update
+      const target = targets?.[targetKey] as TargetConfiguration & {
```
AgentEnder, line 94: Shouldn't need nullish handling here, we know targets[targetKey] exists here.

libs/vscode/nx-workspace/src/lib/get-nx-workspace.ts
```diff
@@ -61,3 +77,28 @@ function getNxCloudEnv(workspacePath: string): any {
...
+        cache?: boolean;
+      };
+
+      if (target?.cache) {
+        cacheableOperations.add(targetKey);
+      }
```
AgentEnder, line 100: `tasksRunnerOptions.default.options.cacheableOperations` is still valid. See here for how we calculate the whole list: https://github.com/nrwl/nx/blob/master/packages/nx/src/tasks-runner/run-command.ts#L501

AgentEnder (dismissed):

libs/vscode/nx-cloud-view/src/lib/nx-cloud-service/nx-cloud-service.ts
MaxKless: This is used to generate some suggestions of commands the user could run to see task runs in nx cloud. I ultimately get the info from going over all targets in the project graph so this should remain accurate even then.

libs/vscode/nx-workspace/src/lib/get-nx-workspace.ts
MaxKless: It's still valid but doesn't necessarily include values. In a new nx 17 workspace there's no more `cacheableOperations` so I need to get that info elsewhere.

AgentEnder: Yeah, but console should check both places right? This won't find any values for existing workspaces or enterprise repos that we don't migrate to the slimmer config.

MaxKless: hmm good point I assumed that the values are populated in the project graph but that was totally unfounded. I can adapt this.

AgentEnder: They won't be in the project graph for a 16.10 or lower repo which we want to still be compatible with.

Issue #1903, opened 2023-10-16
Nx Generate (UI) not working in WebStorm


uus-dev:
## Current Behavior

In the WebStorm IDE I can't use the "Nx Generate (UI)", which opens the UI view to generate libraries, components etc. This has worked before and I can see the "Nx Generate (UI)" button when I right click folders.

Refactor > "Move Nx Project" and Refactor > "Remove Nx Project" still works as well as "Nx: Run Target".

## Expected Behavior

I expect to see the Nx Geneate UI, where you first see a dropdown where you can select what to generate and then after that you can see the full generate window where you can type in the relevant information and choose the relevant options to generate the library, component etc.

## Steps to Reproduce

1. In WebStorm install the "Nx Console" plugin and ensure that the IDE has installed and enabled the plugin
2. Go to any folder and right-click to see the context menu
3. Choose either New > "Nx Generate (UI)" or just directly "Nx Generate (UI)"
4. See that nothing happens

### Failure Logs / Images / Videos

https://github.com/nrwl/nx-console/assets/15202984/be565296-78b4-45a1-85d0-c030cd28dd70

### Environment

- Jetbrains WebStorm version: both 2023.1.4 and 2023.2.3
- Nx Console version: 1.13.1, I also tried to install 1.12.0 and 1.11.1, manually and the experienced the same exact behavior.
- Nx Version:
  - Local: v16.4.0
  - Global: v16.9.1
- `nx report`
  ```
  >  NX   Report complete - copy this into the issue template

   Node   : 18.15.0
   OS     : darwin-arm64
   yarn   : 3.3.1

   nx (global)        : 16.9.1
   nx                 : 16.4.0
   @nx/js             : 16.4.0
   @nx/jest           : 16.4.0
   @nx/linter         : 16.4.0
   @nx/workspace      : 16.4.0
   @nx/angular        : 16.4.0
   @nx/cypress        : 16.4.0
   @nx/devkit         : 16.4.0
   @nx/eslint-plugin  : 16.4.0
   @nx/plugin         : 16.4.0
   @nx/storybook      : 16.4.0
   @nrwl/tao          : 16.4.0
   @nx/webpack        : 16.4.0
   typescript         : 5.1.3
   ---------------------------------------
   Community plugins:
   @fortawesome/angular-fontawesome : 0.13.0
   @ngrx/component                  : 16.0.1
   @ngrx/component-store            : 16.0.1
   @ngrx/effects                    : 16.0.1
[...]

PR #1911, opened 2023-10-19
fix(intellij): make sure jcef is supported before opening

MaxKless:
Surfaces JCEF errors so the users knows what's going on.
image

should fix https://github.com/nrwl/nx-console/issues/1903


iguissouma:
Maybe add a link on the popup:
https://plugins.jetbrains.com/docs/intellij/jcef.html#enabling-jcef

By default it should be enabled since 2020.X
But users can follow this instruction to enable it if they are using a custom jre

image

Cammisuli (approved):

PR #1914, opened 2023-10-20
chore: bump @parcel/watcher to 2.3.0

KwintenP:
👋 
At stackblitz, we have clients wanting to run the extension in codeflow. In order for that to work, we can't directly use @parcel/watcher as a native dependency. Starting from version 2.3.0, @parcel/watcher supports a wasm version. We leverage that and auto-swap out the native dependency for the wasm one when an extension tries to use it (at install time). 

This PR bumps the version to 2.3.0 so its on the version that introduced the wasm package.


MaxKless:
Hey! Sorry this got totally lost - @parcel/watcher is a bit of a contested dep for us and we're not sure how long we're going to keep it. We will make sure to update it though

KwintenP:
That is awesome. 🙏 Do you have an ETA for when this would be released?

MaxKless:
Hey! Super sorry for the long wait, we're testing it now and should be released soon after.

KwintenP:
Hate to ping you again but here I am :). Did the release work? I didn't see it popup in the marketplace and the `package.json` still mentions `2.2.0`.  🙏

MaxKless:
Hey! I tried following you on twitter :) Can you accept so we can chat about it? 
For new nx workspaces, I have a PR to drop @parcel/watcher completely so I'm not sure if this update would solve anything.

MaxKless:
also can you rebase so we can get those pipelines green? :)

MaxKless (approved):

PR #1928, opened 2023-11-07
feat(generate-ui): rework ui with cwd field and put buttons in footer

MaxKless:
WIP PR that modifies the Generate UI.

Download this `.vsix` file and install it using VSCode to test: 
[nx-console.zip](https://github.com/nrwl/nx-console/files/13279867/nx-console.zip)

This zip is for JetBrains IDEs.
Just go to `Settings -> Plugins -> ⚙️` and select `Install Plugin from Disk`
image
[nx-console-0.0.1.zip](https://github.com/nrwl/nx-console/files/13280399/nx-console-0.0.1.zip)

### DIFF
--- a/apps/generate-ui-v2/src/components/badge.ts
+++ b/apps/generate-ui-v2/src/components/badge.ts
@@ -48,7 +48,7 @@ export class Banner extends EditorContext(LitElement) {
     this.dispatchEvent(new CustomEvent('remove'));
   }
 
-  protected createRenderRoot(): Element | ShadowRoot {
+  protected createRenderRoot() {
     return this;
   }
 }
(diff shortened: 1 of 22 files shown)


tbogard:
I like the feature :) one missing thing is that angular services don't have this feature (more likely because we don't have a Nx generator for Angular Services.

Aside of that, thumbs up to this

If curious, here my vod: https://www.twitch.tv/videos/1971541158

Markers:


New
0:29:36: "Angular Service generator not following as-provided"
0:35:33: "only angular project? you can add React too!"
0:40:43: "adding react component with projectNameAndRootFormat=as-provided]"
0:45:25: "generator broke???"
0:52:32: "nx 18 will require full directory path from workspace to any generated component"
0:52:43: "contradictory setup?"
1:04:54: "Angular service what????"
1:08:08: "back to pnpm issue"

PR #1939, opened 2023-11-13
refactor: remove old generate ui

Cammisuli:


Cammisuli:
We'll continue this later

Issue #1959, opened 2023-12-13
Option to sort targets alphabetically


pietdevries94:
## Description

I would like it if the targets were to be sorted alphabetically. An option for this is fine.

## Motivation

We have around 20 targets in a large project and we want to group some targets that do similar things or are related in some way, like seperate, but related steps in our deployment. A prefix (release:foo, release:bar) would be good option for this, but without a way to sort the targets, this would still be chaotic in the vs code extension, where we use `run target` a lot.

## Suggested Implementation

Add an option to sort targets alphabetically in all lists of targets.

## Alternate Implementations

No option and change this. It feels intrusive to other users.

MaxKless:
I like the idea in general :) 
Sorting when we retrieve the targets seems like it would be the easiest solution. However, that would introduce a performance overhead. Especially for large workspaces, iterating over every single project and sorting keys will take time. 
Thus, we should sort targets where they are used to save time for everything else... 
Probably:
- the `Nx: Run Target` and `Nx: run` commands
- The project view
- The `Generate & Run Target` section

Doesn't seem like it would be an intrusive change to me. Right now there's no sorting so it's not like we're breaking any other uses cases...

pietdevries94:
The only case I can think of is that people have made an explicit choice in the order of the targets of their apps (first serve than build for example). Due to the way the targets get retrieved, the result is somewhat predictable. Insert the xkcd about every change breaking a workflow here, haha

PR #1963, opened 2023-12-14
feat(vscode): Sort target names alphabetically

pietdevries94:
Implements #1959 
I checked if it works with run target and in the nx panel. Both are sorted alphabetically.


MaxKless:
Sorry for the slow communication. This looks good, I'll have a look at it soon to see if there's anything we missed.
Appreciate the PR a lot!!

libs/vscode/nx-cli-quickpicks/src/lib/select-run-information.ts
MaxKless, line 1: also sort in the other if condition:

```
if (projectName) {
    return Object.keys(workspace.projects[projectName].targets || {}).sort();
  }
  ```

MaxKless (changes requested):
one sort is missing, otherwise it looks good. 

For the project view, I would not sort the projects for now. Since the project view is kind of a visualisation of project.json files, it makes sense to keep the same order.

But for the run target view & quickpicks, it's a good idea. So just one small change.

Do you have access to IntelliJ? Because we'll need to implement the changes there too.

pietdevries94:
I did the two changes as requested :) 

I sadly don't have an IntelliJ setup and I also haven't written any Java for like 15 years, so I don't feel as comfortable to make the same change at that side.

I however have enabled edits on this PR so if you want to do it, feel free to add it here. And if someone else wants to do it, feel free to ask, I'm fine with inviting people on that repo

PR #2001, opened 2024-01-29
feat(vscode): show codelenses in config files that create targets

MaxKless:
image


libs/vscode/project-details/src/lib/config-file-codelens-provider.ts
```diff
@@ -0,0 +1,293 @@
...
+      return await this.resolveOpenPDVCodeLens(codeLens);
+    } else {
+      return await this.resolveRunTargetCodeLens(codeLens);
+    }
+  }
+
+  private async resolveRunTargetCodeLens(
+    codeLens: RunTargetCodeLens
+  ): Promise<CodeLens> {
+    console.log('project root', codeLens.projectRoot);
```
Cammisuli, line 118: left over debugging?

Cammisuli (approved):

PR #2007, opened 2024-02-01
fix(vscode): improve broken project graph error handling across the board

MaxKless:
- propagate workspace error & reload codelenses
- better error handling
- don't fall back on deprecated readWorkspaceConfig from 17.3 onwards
- PDV preview can start without knowing the project by path

### DIFF
--- a/libs/language-server/watcher/src/lib/watcher.ts
+++ b/libs/language-server/watcher/src/lib/watcher.ts
@@ -22,6 +22,8 @@ const NX_PLUGIN_PATTERNS_TO_WATCH = [
   '**/webpack.config.{js,ts,mjs,cjs}',
   '**/jest.preset.js',
   '**/tsconfig.*.json',
+  // nx-dotnet
+  '*{.csproj,fsproj,vbproj}',
 ];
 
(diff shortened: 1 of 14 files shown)


libs/language-server/watcher/src/lib/watcher.ts
```diff
@@ -6,16 +6,37 @@ import { getNxVersion } from '@nx-console/language-server/workspace';
...
+  '**/playwright.config.{js,ts,cjs,cts,mjs,mts}',
+  '**/remix.config.{js,cjs,mjs}',
+  '**/.storybook/main.{js,ts,mjs,mts,cjs,cts}',
+  '**/{vite,vitest}.config.{js,ts,mjs,mts,cjs,cts}',
+  '**/webpack.config.{js,ts,mjs,cjs}',
+  '**/jest.preset.js',
+  '**/tsconfig.*.json',
+  // nx-dotnet
+  '*{.csproj,fsproj,vbproj}',
+];
```
Cammisuli, line 27: We should get this from the daemon eventually. We should be able to get the plugin globs there. We'll do this later

PR #2011, opened 2024-02-01
fix(nxls): attempt to reconfigure nxls after error with backoff

MaxKless:


Cammisuli (approved):
looks to be working much better now!

Issue #1882, opened 2023-09-13
Lots of symlinks causes webstorm to hang on startup


forivall:
## Current Behavior

In a project with a build process that produces a lot of symlinks, starting WebStorm causes it to hang and consumes 100% CPU (of one core). Disabling the Nx Console plugin or removing the build folder resolves the issue.

Or, click "refresh" on the nx console panel and it will eventually hang, just not immediately.

## Expected Behavior

The Webstorm / IDEA Nx Console plugin should not freeze/crash the IDE.

## Steps to Reproduce

https://github.com/nrwl/nx-examples/pull/296

Run `node ./make-lots-of-symlinks.js`

Start webstorm.

Observe behaviour.

### Failure Logs / Images / Videos

<img width="615" alt="Screenshot 2023-09-12 at 9 00 33 PM" src="https://github.com/nrwl/nx-console/assets/760204/180169be-729c-4703-be00-e93d04238e90">

### Environment

- Nx Console version: 1.13.1
- Webstorm version:
```
WebStorm 2023.2.1
Build #WS-232.9559.58, built on August 22, 2023
Runtime version: 17.0.8+7-b1000.8 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 13.5.2
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 10
Metal Rendering is ON
Non-Bundled Plugins:
    com.github.weisj.darkmode (1.8.0-2023.2)
    nsu.Kirsanov.Nikita (1.2)
    com.intellij.plugins.sublimetextkeymap (231.7515.9)
    com.nasller.CodeGlancePro (1.7.5)
    light.flat.theme (0.14)
    com.sburlyaev.terminal.plugin (0.5.2)
    net.seesharpsoft.intellij.plugins.file-preview (1.6.4)
    de.timo_reymann.intellij-mjml-support (2.4.2)
    dev.nx.console (1.13.1)
    ru.serafimarts.roboticket (1.1.3)
    com.github.menwhorust.tomorrownighttheme (0.0.22)
```

MaxKless:
I'll check it out, appreciate the repro PR :)

PR #2012, opened 2024-02-02
fix(intellij): resolve symlink issue

forivall:
Fixes https://github.com/nrwl/nx-console/issues/1882

This seems to be an underlying issue with intellij's
`isRecursiveOrCircularSymlink` method. An issue and fix should be
submitted upstream.

Until a release of intellij / webstorm / etc. with an upstream fix happens, this solution should help.


MaxKless:
Thanks for the PR! We're a little swamped right now and in the middle of a big release, but I'll test it out and merge it ASAP.

MaxKless (approved):
works like a charm, thank you :)

PR #2022, opened 2024-02-09
fix(intellij): provide Angular config for Angular support in Nx workspaces

piotrtomiak:
Fixes https://youtrack.jetbrains.com/issue/WEB-65265 in Nx projects.

This change requires IntelliJ platform 2023.3.4 as the target. The new version is in RC stage, and is going to be released next week.

I've tested the change in the dev build and with react dependency in `package.json`, HTML files within Angular projects are correctly recognized. However, I am not sure if the sole presence of `project.json` makes it an Angular project - I am not that familiar with Nx workspaces. Let me know if there are some more conditions to be applied here. Please note, that this code cannot use `NxProjectService` as it is run during indexing and must be performance optimized.


lppedd:
I don't think the sole presence of `project.json` makes it an Angular project or library.  
For example, the following is the `project.json` content of one of my non-buildable libraries:

```json
{
  "name": "something-ui-components",
  "$schema": "../../../node_modules/nx/schemas/project-schema.json",
  "sourceRoot": "client-ui/something/components/src",
  "prefix": "something",
  "projectType": "library",
  "tags": ["scope:something:ui"],
  "targets": {
    "test": {
      "executor": "@nx/jest:jest",
      "outputs": ["{workspaceRoot}/coverage/{projectRoot}"],
      "options": {
        "jestConfig": "client-ui/something/components/jest.config.ts"
      }
    },
    "lint": {
      "executor": "@nx/eslint:lint",
      "options": {
        "lintFilePatterns": [
          "{projectRoot}/**/*.ts",
          "{projectRoot}/**/*.html"
        ]
      }
    }
  }
}
```

How do we infer this folder contains Angular components?

MaxKless:
Hmm this is interesting. I agree @lppedd right now the logic doesn't just pick up angular projects. The autocomplete will also show up in a random `.html` file (for example an `index.html` in a react app). 

The only 'real' way I can think of to determine if a project is an angular project if it either
- uses an angular executor / builder like `@nx/angular:browser-esbuild` or `@angular-devkit/build-angular:browser` 
- is depended on by a project that uses an angular builder

This way we could also pick up projects that just contain some angular component without being buildable.
The problem is this would involve constructing the nx project graph via nxls so it's async and not fast during indexing.

Options I can think of:
- check if a `.ts` file imports from `@angular` (involves lots of parsing so expensive and seems like a bad idea)
- check if there's an angular executor / builder in `project.json` (this only works on angular apps, not libs - so it's not fully functional)
- figure out a way to do this async (full correctness, not sure if that's possible though @piotrtomiak)
- leave it like this and live with the fact that people in Nx workspaces will get angular recommendations in other `.html` files. 

Thoughts?

lppedd:
Personal opinion is this is going to be complex, one way or the other.

[WEB-65331](https://youtrack.jetbrains.com/issue/WEB-65331) is what I'd do first, and then I'd transition to a smarter solution using Nx APIs, proven it doesn't take minutes obviously, but I think Nx is fast enough! 

It doesn't have to be strictly at indexing time, as we can just spawn another background process and notify the user when a file is opened for editing.

piotrtomiak:
@MaxKless @lppedd - sorry for the delay! I think we can do some approximate graph build for projects on the IDE side. If not, I can work on the async solution - it is still doable, but results must be stable between calls. Could you please share some example Nx workspace with React, NG component lib and NG project, so that I can see how the dependencies are handled and experiment with it a bit?

Important question is - where `project.json` can be located? Are there some rules for that, or can it be literally at any directory depth?

lppedd:
@piotrtomiak 

> Important question is - where project.json can be located?

It can be present at any depth. And as far as I know, it may also be **not** present, as Nx has inferring capabilities.  
That's why going down the Nx APIs route is still the favorable approach imo.

piotrtomiak:
@lppedd - Yeah, that what I was afraid of. Anyway, I've changed the code to cache the list of projects, so that indexing performance is not affected.

@MaxKless See the inline comments

apps/intellij/build.gradle.kts
```diff
@@ -36,7 +36,14 @@ group = properties("pluginGroup")
 // Configure project's dependencies
 repositories { mavenCentral() }
 
-configurations.all { exclude("org.slf4j", "slf4j-api") }
+configurations.all {
+  exclude("org.slf4j", "slf4j-api")
+  exclude("org.jetbrains.kotlin", "kotlin-stdlib-jdk7")
```
piotrtomiak, line 41: This excludes allow for coroutines compatibility with the platform.

apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularConfigService.kt
```diff
@@ -0,0 +1,111 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import org.angular2.cli.config.AngularConfig
+import org.angular2.cli.config.AngularConfigProvider
+
+@Service(Service.Level.PROJECT)
+@State(name = "NxAngularConfigService", storages = [Storage("nx-angular-config.xml")])
+class NxAngularConfigService(private val project: Project, private val cs: CoroutineScope) :
```
piotrtomiak, line 26: Thanks to compatibility I can get the `couroutineScope` from the platform. It has correct lifecycle and gets cancelled when project or plugin are unloaded. I think you should convert other services to use injected scope instead of the ad hoc ones.

apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularConfigService.kt
```diff
@@ -0,0 +1,109 @@
...
+    PersistentStateComponent<NxAngularConfigService.AngularConfigState> {
+
+    init {
+        project.messageBus
+            .connect(cs)
```
piotrtomiak, line 31: @MaxKless You should also connect to message bus with either a disposable or coroutine scope, so that plugin unloading is handled correctly.

apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularConfigService.kt
```diff
@@ -0,0 +1,109 @@
...
+        val workspaceConfig = workspaceRoot?.findFileByRelativePath("nx.json")
+        if (workspaceRoot == null || workspaceConfig == null) {
+            updateConfig(null)
+            return
+        }
+        val projectFiles =
+            workspace.workspace.projects.values
+                .asSequence()
+                // TODO use dependency graph here, or framework property
+                .filter { project ->
```
piotrtomiak, line 74: @MaxKless So, this is the place where it's up to you now to improve the detection mechanism. Dependency graph is nice, but maybe you could add a property `framework` to the `project.json` spec, so that users can explicitly specify which framework they intend to use and get support for. These properties could be used both by WebStorm and VSCode to determine things like Vue or Angular Language Service startup.

If you add that property I will further improve the code, so that React and Vue frameworks are properly detected within the project.

apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularProject.kt
```diff
@@ -0,0 +1,80 @@
...
+        get() = resolveFile(ngProject.targets?.build?.options?.tsConfig)
+
+    override val karmaConfigFile
+        get() = resolveFile(ngProject.targets?.test?.options?.karmaConfig)
+
+    override val protractorConfigFile
+        get() = resolveFile(ngProject.targets?.e2e?.options?.protractorConfig)
+
+    override fun getTsLintConfigurations(project: Project): List<AngularLintConfiguration> =
+        ngProject.targets?.lint?.let { lint ->
```
piotrtomiak, line 53: I haven't seen any `lint` specific configs, so I guess this whole code may be removed, correct?

apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularProject.kt
```diff
@@ -0,0 +1,80 @@
...
+
+import com.intellij.lang.Language
+import com.intellij.lang.css.CSSLanguage
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VirtualFile
+import org.angular2.cli.config.AngularJsonProject
+import org.angular2.cli.config.AngularLintConfiguration
+import org.angular2.cli.config.AngularProject
+
+class NxAngularProject(
```
piotrtomiak, line 11: This class not only provides context determination, but also some metadata for WebStorm to handle Angular project specifics, like global stylesheets, better.

apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularProject.kt
piotrtomiak: @lppedd - the question is whether the lint config is correctly provided to WebStorm. If yes, and Nx plugin handles that well, we can indeed skip this method and return empty list.

DenisValcke:
I came here from the [WEB-65265](https://youtrack.jetbrains.com/issue/WEB-65265/Angular-support-is-disabled-if-package.json-has-react-in-production-dependencies) issue. Is there a consensus on the fix and will it be released soon? Or should I just downgrade to Webstorm 2023.2.6 and stay there for a little while?

apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularConfigService.kt
MaxKless: Okay, I will build something using the dependency graph here. I don't think a 'framework' property is an option here since it would be such a cross-cutting change. 

Appreciate the work you put in, especially making it async!!

piotrtomiak: > I don't think a 'framework' property is an option here since it would be such a cross-cutting change

I would think of this property as an optional thing to specify, if auto-detection doesn't work. But I don't insist, WebStorm will have it's own solution here as well to facilitate situations when auto-detection fails.

PR #2027, opened 2024-02-20
feat: Implemented eslint problem matcher

work933k:
I propose to have NX-console-extension to support the VSCode problemmatcher functionality for eslint-tasks.


MaxKless (changes requested):
Thanks for the PR! It works great when running tasks from the project view.
However, there's additional places we should consider:
- using the `Nx: run` and `Nx: run target` commands should also populate the problems pane
- running the target from the tooltip in the integrated graph & project details view should also populate the problems pane
- running the target from the `Generate & Run Target` view should also populate the problems pane

I think the easiest solution here is to check for the problemMatchers right inside `CliTaskProvider.executeTask`. You can use the `positional` argument when `command === run`, split it by `:` and check if it's a lint target. In that case, pass the problem matchers to the task.

Thanks a lot for this contribution it rocks!

MaxKless:
Hey :) Am I missing something or are a lot of the places where we could be running tasks from still not integrated here?
We are still not passing problem matchers anywhere except the sidebar when using `CliTaskProvider.instance.executeTask`.

GrumpyMeow:
I wanted to get this PR going again. I thought i had everything working and was finishing up. But then i noticed that when clicking on some "problems" VSCode complained about "non-existant" files. 
I've not came across this before, apparently i got the output from NX-cache which was put there by you (?).  
You probably have your development files in "/home/workflows/workspace/..." while i use a different root-folder...

I triggered affected lint via the NX-console and got this output:
```
Executing task: yarn nx affected --target=lint --output-style=stream 
...
NX   Successfully ran target lint for 10 projects
Nx read the output from the cache instead of running the command for 10 out of 10 tasks.
```
A lint-result was:
...
/home/workflows/workspace/libs/language-server/utils/src/lib/runtime.ts
   4:16  warning  'configureHttpRequests' is defined but never used                             @typescript-eslint/no-unused-vars
  17:29  warning  '_encoding' is defined but never used. Allowed unused args must match /^_$/u  @typescript-eslint/no-unused-vars

✖ 2 problems (0 errors, 2 warnings)
```
This should be a very common issue with NX-tasks which output absolute-filepaths (in contrary to relative paths). 

I could probably try to see if there is a ESLint-reporter/format which outputs relative-filepaths.  

Weird right?  Any thoughts?

work933k:
I'm gonna create a new PR..

PR #2037, opened 2024-02-29
fix(intellij): fix issue with NxAngularConfigService state not being stored

piotrtomiak:


MaxKless (approved):
works great thanks :)

PR #2102, opened 2024-04-23
chore: set up @nx/cypress/plugin

leosvelperez:
Sets up `@nx/cypress/plugin`:

- with task splitting
- keeping the exact same options for the `e2e` task
- the `watch` configuration becomes the `open-cypress` task


leosvelperez:
@MaxKless let me know if you want me to change CI to use the `e2e-ci` task instead.

MaxKless (approved):

PR #2115, opened 2024-05-02
feat: enable partial graphs in nx console & fix bugs

MaxKless:
- catch errors & use partial graph
- rework graph error display & fixes
- call nx daemon --stop instead of nx reset on refresh
- various fixes
- formatting
- updated error display
- handle broken nx.json
- further intellij fixes
- fix run-gradle.js again
- setColors on wrapped browser load html
- more intellij fixes
- enable partial graph display in nx 19 beta

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/graph/NxGraphBrowserBase.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/graph/NxGraphBrowserBase.kt
@@ -377,12 +377,12 @@ abstract class NxGraphBrowserBase(protected val project: Project) : Disposable {
               const darkClass = 'vscode-dark';
               const lightClass = 'vscode-light';
 
-              body.classList.remove(darkClass, lightClass);
+              body.classList?.remove(darkClass, lightClass);
 
               if (isDark) {
-                  body.classList.add(darkClass);
+                  body.classList?.add(darkClass);
               } else {
-                  body.classList.add(lightClass);
+                  body.classList?.add(lightClass);
               }
               console.log("$backgroundColor")
               body.style.setProperty('background-color', '$backgroundColor', 'important');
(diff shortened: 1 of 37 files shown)


apps/intellij/src/main/kotlin/dev/nx/console/graph/NxGraphBrowserBase.kt
```diff
@@ -321,8 +325,9 @@ abstract class NxGraphBrowserBase(protected val project: Project) : Disposable {
               }
             </style>
 
-            <p>Unable to load the project graph. The following error occurred:</p>
-      <pre>${error}</pre>
+            <p>Unable to load the project graph. The following error${if(errors.isNotEmpty()) "s" else ""} occurred:</p>
+      ${errors.map { "<pre>${it.message ?: ""} \n ${it.stack ?: ""}</pre>" }.joinToString("\n")
+        }
       If you are unable to resolve this issue, click here to <a href="#" onclick="window.reset()">reset</a> the graph.
```
jaysoo, line 331: Should we also say to run `nx reset` in the terminal as an alternative? In the case of the `CancellationException` it is the only way to resolve.

apps/intellij/src/main/kotlin/dev/nx/console/nxls/NxlsProcess.kt
```diff
@@ -58,7 +57,7 @@ class NxlsProcess(private val project: Project, private val cs: CoroutineScope)
 
                             logger.trace("Nxls early exit: $this")
 
-                            Notifier.notifyNxlsError(project)
+                            //                            Notifier.notifyNxlsError(project)
```
jaysoo, line 60: We don't want this notifier?

MaxKless: No, the exit is now triggered in "allowed" circumstances when restarting. We will need to provide a message when the nxls errors out in the future, but this was being triggered wrongly.

jaysoo: Let's remove it then

jaysoo (approved):

PR #2131, opened 2024-05-09
feat(intellij): autofocus first empty text field in re/move dialog

mkapal:
The project text field is focused by default when the re/move project dialog is opened. This improves UX because the user doesn't have to move the focus to the project field manually.

Before:
Screenshot 2024-05-09 at 20 35 03

After:
Screenshot 2024-05-09 at 20 35 51

When the project is prefilled then the destination text box is focused instead:
image


MaxKless:
I like it, thanks for the PR!
What do you think about the following: If the project field is prefilled (for example when you right click on a folder & run it via the refactoring menu there), the destination field should be focused instead.

mkapal:
I just noticed that I import the `applyIf` function from `ai.grazie.utils.applyIf` - somehow that doesn't look like the right way to do it.

Is there a better standard way to apply such properties based on a condition?

mkapal:
Never mind, now I see that it is also available in `com.intellij.util.applyIf`. That's fixed now.

MaxKless (approved):

MaxKless:
looks & works great!
Can you run `./gradlew ktfmtFormat` to make sure linting passes @mkapal ?

PR #2134, opened 2024-05-10
chore: use windows agents

MaxKless:
- chore: migrate nx cloud to staging
- chore: use read-write token only in CI
- fix ci targets
- change syntax again
- try fixing windows syntax
- remove unused intellij test
- try using windows agents
- set NX_CI_EXECUTION_ENV


apps/nxls-e2e/project.json
```diff
@@ -5,16 +5,7 @@
...
-      "options": {
-        "cwd": "apps/nxls-e2e"
-      },
-      "dependsOn": [
-        {
-          "projects": "nxls",
-          "target": "build"
-        }
-      ]
+      "executor": "nx:noop"
```
leosvelperez, line 8: This shouldn't be needed. I can see you already have the plugin configured to infer an `e2e` target and not a `test` target. We should be able to delete the `test` target here.

leosvelperez (approved):

PR #2146, opened 2024-05-22
chore: add unit tests for generator schema utils

mkapal:
I thought it would be nice to have common functions covered by unit tests, to prevent regression bugs in the future.

I have started with normalize schema utils in `apps/generate-ui-v2/src/utils`.

If you think this is a good way to go, I can go and analyze which functions are used the most / are most critical and cover them with tests. I'm also open for suggestions what to look at first.

### DIFF
--- a/apps/generate-ui-v2/src/utils/generator-schema-utils.spec.ts
+++ b/apps/generate-ui-v2/src/utils/generator-schema-utils.spec.ts
@@ -1,0 +1,174 @@
+import {
+  compareWithDefaultValue,
+  extractDefaultValue,
+  extractItemOptions,
+  getGeneratorIdentifier,
+  getGeneratorNameTitleCase,
+} from './generator-schema-utils';
+import { OptionType } from '@nx-console/shared/schema';
+
+describe('generator schema utils', () => {
+  describe('getGeneratorIdentifier', () => {
+    it('should return an empty string if the generator schema is undefined', () => {
+      expect(getGeneratorIdentifier(undefined)).toEqual('');
+    });
+
+    it('should return the generator identifier', () => {
+      expect(
+        getGeneratorIdentifier({
+          generatorName: 'library',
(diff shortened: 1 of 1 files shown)


mkapal:
@MaxKless What are your thoughts on this?

MaxKless:
Hey! Thank you so much @mkapal I think you're right and it's a great idea!! 

The tests you have already written look great, I will approve the PR and you can just merge it like this too and do addtl things in a follow up PR - up to you. 

Regarding future tests: 
- I think if you want to stay in the same area, `apps/generate-ui-v2/src/form-values.service.ts` is a very important service with lots of behavior that could be tested. It will require some mocking though.
- `libs/shared/utils` is also something that could be good to start with. A lot of things in Nx Console interact with the node filesystem API though, so that will always have to be mocked in some way. It might require some work.
- Something I've started recently which I think is **very** important is nxls e2e tests under `apps/nxls-e2e`. They take the full spectrum of `nx` and Nx Console into consideration and can work between different versions. If you're willing to get into it, I would love some help! Basically every file in `libs/language-server/workspace/src/lib` needs a corresponding test. I would also love to hear your feedback on the API I've written for the e2e tests, how does it feel to use and what would you improve?

Just want to say again that I really appreciate your help :)

MaxKless (approved):

mkapal:
I agree, let's merge this PR as it is and I can continue with the test coverage and will create follow-up PRs.

Thanks for the test suggestions. I might start with something simpler as i still need to wrap my head around the repo, especially the nxls package.

I'm using the Nx Console plugin myself in WebStorm and it's helpful, so I'm happy to contribute to the project. 👍

MaxKless:
ok sounds good! If you need any help or have questions about the repo feel free to ping me here or on twitter or maxk@nrwl.io :) 
I'd be glad to support any contributions!

PR #2152, opened 2024-05-28
fix: `npmDependencies` returns dependencies from `.nx/installation/node_modules` if `node_modules` is a file

mkapal:
When writing unit tests for [`npmDependencies`](https://github.com/nrwl/nx-console/blob/bfd3c3d4ddb37c8a5fcb6c4b3a3445feb0b3fce5/libs/shared/npm/src/lib/npm-dependencies.ts), I decided to replace `await stat(x)).isDirectory()` with [`directoryExists(x)`](https://github.com/nrwl/nx-console/blob/bfd3c3d4ddb37c8a5fcb6c4b3a3445feb0b3fce5/libs/shared/file-system/src/lib/directory-exists.ts) from `@nx-console/shared/file-system` and added a `readDirectory` function to make it easier to mock those file system operations.

Since the `directoryExists` function already handles thrown exceptions inside, I could refactor the main code by getting rid of the nested try...catch blocks, but then I noticed a strange code path: Before, if `node_modules` was not a directory but an existing file, an empty array was returned regardless whether `.nx/installation/node_modules` directory existed. Now it goes on to check if `.nx/installation/node_modules` exists too. I think this is the correct behavior now, albeit quite an unlikely scenario.

You can check the new test cases to see if my understanding is correct.

### DIFF
--- a/libs/shared/file-system/src/index.ts
+++ b/libs/shared/file-system/src/index.ts
@@ -5,4 +5,5 @@ export { readAndCacheJsonFile } from './lib/cache-json';
 export { clearJsonCache } from './lib/cache-json';
 export { readAndParseJson } from './lib/cache-json';
 export { listFiles } from './lib/list-files';
+export { readDirectory } from './lib/read-directory';
 export { readFile } from './lib/read-file';
--- a/libs/shared/file-system/src/lib/read-directory.ts
+++ b/libs/shared/file-system/src/lib/read-directory.ts
@@ -1,0 +1,9 @@
+import { readdir } from 'fs/promises';
+
+export async function readDirectory(path: string): Promise<string[]> {
+  try {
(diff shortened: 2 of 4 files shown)


MaxKless (changes requested):
This looks great, thanks for the refactor and the tests @mkapal
Since we have tests now, I think let's add another case: There is some logic for handling nested dependencies that start with `@`. Can you add another test case for that code path? 
Thank you!

mkapal:
Good idea, I've added that now: https://github.com/nrwl/nx-console/blob/94f9318f50bc2213cc725d71bc90644fdf3a8a15/libs/shared/npm/src/lib/npm-dependencies.spec.ts#L84

MaxKless (approved):
looks great, thanks so much!

PR #2175, opened 2024-06-19
feat(intellij): add cloud view & add project graph errors to problems pane 

MaxKless:
- create cloud view lib
- changes
- flow with connect works
- open correct cloud url
- add e2e tests


apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt
```diff
@@ -0,0 +1,96 @@
...
+import dev.nx.console.nxls.NxWorkspaceRefreshListener
+import dev.nx.console.nxls.NxlsService
+import dev.nx.console.utils.nxBasePath
+import java.io.File
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+@Service(Service.Level.PROJECT)
+class ProjectGraphErrorProblemProvider(val project: Project, val cs: CoroutineScope) {
+    private val nxProblemsProvider = NxProblemsProvider(project)
```
jaysoo, line 24: What is this for? I don't see any usage of `nxProblemsProvider`. It's just for `FileProblem`? Does it need the override there? A comment explaining would be helpful.

apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt
```diff
@@ -0,0 +1,96 @@
...
+    private val nxlsService = NxlsService.getInstance(project)
+
+    private val problems = mutableListOf<Problem>()
+
+    fun init() {
+        setNxProblems()
+        with(project.messageBus.connect()) {
```
jaysoo, line 31: Nit: Is this the same as

```
project.messageBus
  .connect()
  .subscribe(
    NxlsService.NX_WORKSPACE_REFRESH_TOPIC,
    NxWorkspaceRefreshListener { setNxProblems() }
  )
```

?

apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    private fun setNxProblems() {
+        cs.launch {
+            val nxWorkspace = nxlsService.workspace()
+            val newProblems =
+                nxWorkspace?.errors?.let {
+                    nxWorkspace.errors.mapNotNull { error -> nxErrorToFileProblem(error) }
+                }
+
+            problems.forEach { ProblemsCollector.getInstance(project).problemDisappeared(it) }
```
jaysoo, line 47: Nit: Could also go to a class member right?

```
private val problemsCollector = ProblemsCollector.getInstance(project)
```
```

File: apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt (hunk 1 above)
Comment (MaxKless) [reply apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt:24]: It's just there to be passed to `FileProblem`. `override` just specifies that it's a property that exists in the interface, so it's required. 
You mean a comment above `nxProblemsProvider`? Like `this is required by FileProblem` or something alike? Just making sure I understand you correctly :)

File: apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt (hunk 2 above)
Comment (MaxKless) [reply apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt:31]: Yeah it's the same. The `with` syntax is useful when subscribing to multiple topics but useless here.

File: apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt (hunk 1 above)
Comment (jaysoo) [reply apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt:24]: Because the class the empty, I thought it isn't needed, but maybe just a `// Empty class because FileProblem requires override` to explain why there's an empty class.

File: apps/intellij/src/main/kotlin/dev/nx/console/nx_toolwindow/NxToolWindowPanel.kt [hunk 4]
```diff
@@ -1,93 +1,231 @@
...
+                    button("View Errors") { ProblemsView.getToolWindow(project)?.show() }
                         .align(Align.CENTER)
                 }
                 row {
                     text(
-                        "If you're just getting started with Nx, you can <a href='https://nx.dev/plugin-features/use-code-generators'>use generators</a> to quickly scaffold new projects or <a href='https://nx.dev/reference/project-configuration'>add them manually</a>." +
-                            "<br/> If your Nx workspace is not at the root of the opened project, make sure to set the <a href='open-setting'>workspace path setting</a>."
+                        "If the problems persist, you can try running <code>nx reset</code> to clear all caches and then <a href='refresh'>refresh the workspace</a><br /> For more information, look for errors in <a href='open-idea-log'>idea.log</a> and refer to the <a href='https://nx.dev/troubleshooting/troubleshoot-nx-install-issues?utm_source=nxconsole'>Nx Troubleshooting Guide </a> and the <a href='https://nx.dev/recipes/nx-console/console-troubleshooting?utm_source=nxconsole'>Nx Console Troubleshooting Guide</a>."
```
jaysoo, line 202: I don't think we need to say `to clear all caches`, just `try running nx reset and then refresh...`.

jaysoo (approved):

apps/intellij/src/main/kotlin/dev/nx/console/ide/ProjectGraphErrorProblemProvider.kt
MaxKless: Yeah I think it's mostly there to hold the project honestly... I can add a comment

PR #2182, opened 2024-06-21
feat(vscode): make generate ui default in sidebar

MaxKless:
Modifies 'Common Nx Commands' so that the "Generate (UI)" command is the default being run
image

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -198,8 +198,8 @@
           "group": "inline@1"
         },
         {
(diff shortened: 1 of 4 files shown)


leosvelperez (approved):

moatorres:
Thank you! Having Generate UI available by default means that this killer feature will continue to blown new-users' heads on first contact. It's such an important part of what NX has to offer. Looking forward to the next release.

PR #2183, opened 2024-06-21
feat(misc): add support for run-help event

jaysoo:
We are showing `--help` content from PDV, which requires that this enhancement lands in Console first, or else `run-help` will do nothing.

See this for more explanation: https://www.loom.com/share/12a5e50762424ad9bb86dbf002e1116a


apps/intellij/src/main/kotlin/dev/nx/console/run/runGeneralCommand.kt
```diff
@@ -0,0 +1,64 @@
...
+
+                nxProject?.root?.also { withWorkDirectory(File(project.nxBasePath).resolve(it)) }
+            }
+
+        val processHandler = KillableColoredProcessHandler(commandLine)
+        val console = TextConsoleBuilderFactory.getInstance().createBuilder(project).console
+
+        console.attachToProcess(processHandler)
+
+        SwingUtilities.invokeLater {
```
MaxKless, line 47: We should use `ApplicationManager.invokeLater` instead of `SwingUtilities`. [See docs here](https://plugins.jetbrains.com/docs/intellij/general-threading-rules.html#modality-and-invokelater).

Since we're in a Coroutine though, we can make this even easier and just use `withContext(Dispatchers.EDT) { ...`. [See docs here](https://plugins.jetbrains.com/docs/intellij/coroutine-dispatchers.html#edt-dispatcher)

libs/vscode/graph-base/src/lib/handle-graph-interaction-event.ts
```diff
@@ -43,5 +41,37 @@ export async function handleGraphInteractionEventBase(event: {
     return true;
   }
 
+  if (event.type === 'run-help') {
+    getTelemetry().featureUsed('nx.graph.runHelp');
+    const workspacePath = getNxWorkspacePath();
+    console.log('>>>', event.payload);
```
MaxKless, line 47: leftover log?

apps/intellij/src/main/kotlin/dev/nx/console/run/runGeneralCommand.kt
```diff
@@ -0,0 +1,64 @@
...
+import com.intellij.openapi.project.Project
+import dev.nx.console.NxIcons
+import dev.nx.console.nxls.NxlsService
+import dev.nx.console.utils.ProjectLevelCoroutineHolderService
+import dev.nx.console.utils.nxBasePath
+import java.io.File
+import javax.swing.SwingUtilities
+import kotlinx.coroutines.launch
+
+fun runGeneralCommand(project: Project, projectName: String, helpCommand: String) {
```
MaxKless, line 19: This doesn't really seem like a `runGeneralCommand` implemention but more like a `runHelpCommand` implementation right? I think variable & function naming should be made consistent either way.

MaxKless, line 19: It works fine like this but we could also think about making this a `NxRunHelpService` or something like that. It's really simple with a `@Service(Service.Level.PROJECT)` annotation. [Read more here](https://plugins.jetbrains.com/docs/intellij/plugin-services.html#light-services).
In that case we would automatically get our own coroutine scope instead of having to use the `ProjectLevelCoroutineHolderService` (which would make debugging easier in case anything goes wrong). Also, we wouldn't have to pass `project` to the function - instead to the service's `getInstance` method though. 
So it doesn't make the biggest of differences but it seems to be the slightly more "intellij-y" way of doing things.

MaxKless (changes requested):
looks good already, just a few tweaks

apps/intellij/src/main/kotlin/dev/nx/console/run/runGeneralCommand.kt
jaysoo: Yeah the intention was originally running any command, but it is more specific now.

MaxKless (approved):

PR #2224, opened 2024-07-30
fix: pass expanded externalFiles to ts plugin for wildcard pattern paths

MaxKless:


Tallyb:
I think this is still not catering to the option where paths have multiple entries. this is a valid paths definition: 
```
@foo: [packages/foo/*, tests/foo/*]
```

libs/vscode/typescript-plugin/src/lib/typescript-plugin.ts
```diff
@@ -133,14 +134,23 @@ async function getExternalFiles(
...
     const configFilePath = await findConfig(mainFile, TSCONFIG_LIB);
 
     if (!configFilePath) {
       continue;
     }
 
-    const directory = dirname(configFilePath);
-    externals.push({ mainFile, directory });
+    if (mainFile.endsWith('/*')) {
+      const files = listFiles(dirname(mainFile));
```
jaysoo, line 145: I don't think nested paths would work.

e.g. If `@acme/foo/*` points to `libs/foo/src`, and someone imports `@acme/foo/nested/module`.

jaysoo (approved):

PR #2242, opened 2024-08-27
fix(vscode): fix behaviour when only 1 generator is available & refactor old code

MaxKless:
- lets user select generator when there is only one generator instead of automatically selecting it (this won't affect the move/remove actions)
- fixes `generatorName` data that is used in generate ui header 
- show proper error when no move / remove generators are available
- simplifies & refactors old code (parts of this were still left over from when there was still a task & generate ui)


jaysoo (approved):
LGTM

PR #2260, opened 2024-09-19
feat(intellij): add Mark Directory As Nx Workspace Root action

MaxKless:


jaysoo (approved):

lppedd:
I suppose this is useful only when the workspace is nested inside another project, right?

MaxKless:
yes exactly, that's the use case :) If the nx workspace isn't at the root of the folder opened in intellij

PR #2272, opened 2024-09-30
feat(vscode): rework PDV integration to use new apis & state machine

MaxKless:
- WIP
- more xstate WIP
- update to xstate 5
- more xstate migration fixes
- state machine start for pdv
- render multi pdv with select box
- multi works
- changes
- fix bugs
- more fixes


libs/vscode/project-graph/src/lib/legacy-implementation/graph.machine.ts
```diff
@@ -27,168 +34,165 @@ interface Context {
...
+        };
+      }),
+      contentLoaded: assign((context, event) => {
+        return {
+          state: State.loaded,
+          error: null,
+        };
+      }),
+      loadingFailed: assign(({ context, event }) => {
+        if (event.type !== 'error.platform.loadingContent') {
```
jaysoo, line 159: Why do we need this guard clause now?

libs/vscode/project-graph/src/lib/legacy-implementation/graph.machine.ts
```diff
@@ -27,168 +34,165 @@ interface Context {
...
+          state: State.loading,
+        };
+      }),
+      clearProject: assign((context, event) => {
+        return {
+          project: null,
+        };
+      }),
+      projectSelected: assign(({ context, event }) => {
+        if (event.type !== 'PROJECT_SELECTED') {
```
jaysoo, line 178: Same here, why is the guard clause needed?

libs/vscode/project-details/src/lib/new-project-details-preview.ts
```diff
@@ -0,0 +1,560 @@
...
+  constructor(private path: string, private context: ExtensionContext) {
+    this.webviewPanel = window.createWebviewPanel(
+      'nx-console-project-details',
+      `Project Details`,
+      ViewColumn.Beside,
+      {
+        enableScripts: true,
+      }
+    );
+    const machine = setup({
```
jaysoo, line 39: Not really an XState expert, but it feels like the render side-effects aren't part of the workflow. Maybe instead it should be a handler to `actor.subscribe` that renders the webview depending on the current state?

libs/vscode/project-details/src/lib/new-project-details-preview.ts
```diff
@@ -0,0 +1,560 @@
...
+          multiSelectedProject: string | undefined;
+        },
+      },
+      actors: {
+        loadPDVData: fromPromise(async () => {
+          return await getPDVData(path);
+        }),
+      },
+      actions: {
+        renderLoading: () => this.renderLoading(),
```
jaysoo, line 51: Not really an XState expert, but it feels like the render side-effects aren't part of the workflow. Maybe instead it should be a handler to `actor.subscribe` that renders the webview depending on the current state?

libs/vscode/nx-workspace/src/lib/get-pdv-data.ts
```diff
@@ -0,0 +1,9 @@
+import { NxPDVDataRequest } from '@nx-console/language-server/types';
+import { PDVData } from '@nx-console/shared/types';
+import { sendRequest } from '@nx-console/vscode/lsp-client';
+
+export async function getPDVData(
```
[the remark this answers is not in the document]
MaxKless: This is a pattern across Nx Console and I don't want to break it in this PR. 
You definitely have a point though, there isn't a huge amount of value gained from this specific function. However, some of the other pdv request functions have different signatures or include some sort of mapping of inputs or outputs so there is more value to them. 
I think it isn't very harmful to have this layer between things so that the vscode extensions doesn't have to concern itself with where the data come from.

libs/vscode/project-graph/src/lib/legacy-implementation/graph.machine.ts
MaxKless: I honestly have no idea. It seems like we could remove it but there could be something I'm missing. My goal was to not mess with the legacy implementation code so I'm not sure I want to change things here.

libs/vscode/project-details/src/lib/new-project-details-preview.ts
MaxKless: Not an XState expert either, so you might be completely right. My thought process goes like this:
While the simple rendering cases (loading, error) could easily be done with a state subscription, the PDV cases with their child states & `reRenderPDV` logic seem more complicated. So in the subscription we would have to branch depending on the previous state of the machine: is it just entering the `showingPDV.idle` state or was it in `showingPDV.refreshing` before? 
So while we would remove the actions from the machine, we would essentially write a subscription that is tightly coupled to the machine implementation, which seems like it's just spreading out the logic instead of actually decoupling it.

jaysoo (approved):

PR #2274, opened 2024-10-01
feat(vscode): add nxWorkspacePath configuration option

### LINKED ISSUE #2268: VSCode Select Workspace should become a workspace-level setting to be shared between devs & sessions
We want the option to select the nx workspace as a nested folder in the open vscode workspace to work just like it does in intellij. 
Users should be able to select a folder and we save the path relative to the currently open workspace. 
That way the setting can be shared between users & sessions.

gabynevada:
This pull request introduces a new configuration option for specifying the Nx workspace path in the VS Code extension. The changes include updates to the configuration schema, new imports and functions to handle the workspace path, and modifications to existing functions to integrate the new setting.

Closes #2268

### Configuration Enhancements:
* [`apps/vscode/package.json`](diffhunk://#diff-ffd752c4d60ebba40a59cc7f9b8cb7b3beeedb5c14a8a8245246bf73846bf247R709-R712): Added a new configuration option `nxConsole.nxWorkspacePath` to specify the relative path to the root directory of the Nx workspace.

### Codebase Updates:
* `apps/vscode/src/main.ts`: 
  - Added new imports `relative` and `resolve` from the `path` module.
  - Integrated `registerSettingsNxWorkspacePathWatcher` in the `activate` function to watch for changes in the new configuration setting.
  - Updated `manuallySelectWorkspaceDefinition` to store the relative path of the selected directory in the global configuration.
  - Modified `scanForWorkspace` to use the new configuration setting if available.
  - Introduced `registerSettingsNxWorkspacePathWatcher` to handle changes in the `nxWorkspacePath` configuration setting.

### Configuration Keys:
* `libs/vscode/configuration/src/lib/configuration-keys.ts`: 
  - Added `nxWorkspacePath` to the global configuration keys.
  - Updated the `GlobalConfig` type to include `nxWorkspacePath`.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -709,7 +709,7 @@
         },
         "nxConsole.nxWorkspacePath": {
           "type": "string",
-          "description": "Specifies the relative path to the root directory of the Nx workspace."
+          "description": "Specifies the path to the root directory of the Nx workspace."
         }
       }
     },
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -706,6 +706,10 @@
           "scope": "application",
           "default": false,
           "description": "Show a notification with the current Node version on startup - useful for debugging nvm issues."
+        },
+        "nxConsole.nxWorkspacePath": {
+          "type": "string",
+          "description": "Specifies the relative path to the root directory of the Nx workspace."
         }
       }
     },
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -709,7 +709,7 @@
         },
         "nxConsole.nxWorkspacePath": {
           "type": "string",
-          "description": "Specifies the path to the root directory of the Nx workspace."
+          "description": "Specifies the relative path to the root directory of the Nx workspace. Can be configured on a user or workspace level."
         }
       }
     },
--- a/apps/vscode/src/main.ts
+++ b/apps/vscode/src/main.ts
@@ -1,5 +1,5 @@
 import { existsSync } from 'fs';
-import { dirname, join, parse, resolve } from 'path';
+import { dirname, join, parse, relative, resolve } from 'path';
 import {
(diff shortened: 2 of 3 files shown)


gabynevada:
https://github.com/user-attachments/assets/85eefd81-7917-4036-98b4-43bbc8328409

MaxKless (changes requested):
Nice! Wow you're fast. Thanks for this.

This is still missing some important pieces imo:
- When using the `Select Workspace` command (`nxConsole.selectWorkspaceManually`, which ends up calling the `main.ts#manuallySelectWorkspaceDefinition` function), the setting should be updated to contain the relative path. 
- There should be a change listener (`workspace.onDidChangeConfiguration`) so that when this setting is changed, `setWorkspace` is called and Nx Console updates automatically. That way users won't have to restart the ide
- [Make sure the configuration is workspace-level](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) (by using the scope property) so it can be shared between anyone who clones the repo. Because realistically, anyone with a nested repo will probably set this setting the same.
- The configuration description should specify that this should be a relative path and is a workspace-level setting

gabynevada:
Makes sense, thanks for the quick feedback!

> * [Make sure the configuration is workspace-level](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) (by using the scope property) so it can be shared between anyone who clones the repo. Because realistically, anyone with a nested repo will probably set this setting the same.

What scope would be the correct one for this? When I was testing, went with the window setting which works in both the workspace and user settings, and since it was the default I removed the property.

gabynevada:
@MaxKless had a busy week from work but pending tasks should be completed now. Let me know if there's any issues or additional changes.

MaxKless (approved):
Thanks, code looks great and works great!

PR #2277, opened 2024-10-07
feat: Implemented VSCode lint problem matching

GrumpyMeow:
With this PR i propose to have VSCode automatically take the output of NX-linting tasks and present the issues as VSCode problems.
This PR implements the matching at:
* Running "affected" tasks
* Running "many" tasks
* Running an individual task
When running affected/many the output-style is automatically changed to display the found issues and allow for matching.

A known issue is that the default ESLint-formatter which is used by NX will report with absolute filepaths. The absolute filepaths might not align with the current environment as NX-caching does not take the root-folder in account.

In the future this feature might be extended to also work for Stylelint and others.

@MaxKless Would you like to check this PR?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,10 @@ Use the `F5` key or the debug menu option `Launch Client + Server` to start the 
 
 > :warning: Even though builds will be generated automatically, the Extension Development Host needs to be restarted in order to apply a new set of changes.
 
+> :warning: Set `window.openFilesInNewWindow` to "off" to prevent VSCode to reopen a new window
+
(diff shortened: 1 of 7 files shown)


MaxKless:
Can you elaborate on a few things?
- Why did you set `--output-style=stream` in a few cases?
- Why if the lint targets are called something other than `lint`? 
- What if the lint targets use a technology other than `eslint`? 
- What happens if the `eslint-stylish` problem matcher isn't installed? Do we run into errors?

We should probably have another solution that checks the executor or makes sure that the targets where created by the `@nx/eslint` plugin to make sure we don't cause any issues.

PR #2283, opened 2024-10-14
refactor(vscode): update projects view to only make a single nxls request

MaxKless:


libs/vscode/nx-project-view/src/lib/views/nx-project-base-view.ts
```diff
@@ -172,7 +172,9 @@ export abstract class BaseView {
   ): Promise<TargetViewItem[] | undefined> {
     const { nxProject, nxTarget } = parent;
 
-    const projectDef = (await getNxWorkspaceProjects())[nxProject.project];
+    const projectDef =
```
jaysoo, line 175: We can wrap this in a method, just to ensure that the workspace data is available. e.g. `getProjectData` or something

jaysoo (approved):

PR #2288, opened 2024-10-17
fix: kill entire process tree when shutting down nxls & vscode

MaxKless:
when closing vscode, we want to make sure all resources (like the language server, daemon, file watchers & the extension process itself are cleaned up).
VSCode exposes a `deactivate` function which we implement. However, there are some limitations to it:
- some APIs may break down during deactivation
- there is a hard limit of 5 seconds before the process is killed forcefully
source: https://github.com/microsoft/vscode/issues/47881#issuecomment-381910587

Especially the second limitation can cause troubles. Since we used to first shut down the daemon and then delegate cleanup to the language server, if one of those steps took longer than 5 seconds, Nx Console would exit without properly cleaning up all of it's resources.
In addition, due to the deeply nested nature of subprocesses that nx spawns, in combination with windows quirks (especially the wonky `windowsHide` handling that breaks `SIGINT` as a concept), simply killing the process tree often didn't catch all processes. 

This PR fixes this:
- We use `taskkill` on windows to circumvent `windowsHide` issues and the like.
- in VSCode, we only send a `shutdown` request to the nxls which kills the daemon and unregisters the file watcher. Actual process termination happens on the extension side
- We do not delegate cleanup to the language server anymore. While this is nice and reusable in theory, it went wrong too often in practice and there isn't much that's happening in cleanup anyways. Now, we simply kill the entire process tree of the nxls, run some shutdown steps and then kill the entire process tree of the extension host


jaysoo (approved):

libs/shared/utils/src/lib/utils.ts
```diff
@@ -80,3 +82,18 @@ export function withTimeout<T>(asyncFn: () => Promise<T>, timeoutMs: number) {
       });
   });
 }
+
+export function killProcessTree(pid: number, signal = 'SIGTERM') {
+  if (process.platform !== 'win32') {
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      require('tree-kill')(pid, signal);
```
jaysoo, line 90: Note that `tree-kill` already uses `taskkill` for Windows. Should maybe look into inlining the function like we did for Nx since that package hasn't been maintained in a long time.

MaxKless: sure, I can do that.

PR #2291, opened 2024-10-22
fix(nxls): adjust parseTargetString invocation to always pass a project graph

MaxKless:
- fix(nxls): adjust parseTargetString invocation to always pass a project graph
- add e2e test and fall back to targets for link


libs/language-server/capabilities/document-links/src/lib/target-link.ts
```diff
@@ -41,20 +48,39 @@ export async function targetLink(
...
     const importPath = join(devkitPath, 'src/executors/parse-target-string');
     const { parseTargetString } = await importWorkspaceDependency<
       typeof import('@nx/devkit/src/executors/parse-target-string')
     >(importPath, lspLogger);
-    const parsedTargets = parseTargetString(targetString);
-    project = parsedTargets.project;
-    target = parsedTargets.target;
-    configuration = parsedTargets.configuration;
+    let parsedTarget: Target;
+    if (gte(nxVersion.full, '17.0.6')) {
```
Coly010, line 58: `semver.gte` will throw if `nxVersion.full` is not a valid `semver` string. 

Do we want to add a `try-catch` around this to prevent crashes and have some kind of fallback? It might never hit it, I'm not sure the order of operations here, but a thought at least.

package.json
```diff
@@ -51,19 +51,19 @@
...
-    "@nx/eslint-plugin": "19.8.0",
-    "@nx/gradle": "19.8.0",
-    "@nx/jest": "19.8.0",
-    "@nx/js": "19.8.0",
-    "@nx/node": "19.8.0",
-    "@nx/storybook": "19.8.0",
-    "@nx/web": "19.8.0",
-    "@nx/webpack": "19.8.0",
-    "@nx/workspace": "19.8.0",
+    "@nx/cypress": "20.0.2",
```
Coly010, line 54: latest Nx is currently `20.0.3`, do we want to use that version instead?

Coly010 (approved):
few comments, but lgtm

libs/language-server/capabilities/document-links/src/lib/target-link.ts
MaxKless: it's a valid point, I'll leave it for now but I have a refactor planned sometime in the future for the whole nxVersion stuff. We need to break out from semver by treating `0.0.0-pr` versions as `999` or smth otherwise nx console is just broken on pr releases

PR #2300, opened 2024-10-30
fix(vscode): refactor new pdv to separate webview & state machine

MaxKless:


libs/vscode/project-details/src/lib/pdv-state-machine.ts
```diff
@@ -0,0 +1,230 @@
...
+    context: {} as Partial<PDVData> & {
+      multiSelectedProject: string | undefined;
+    },
+  },
+  actors: {
+    loadPDVData: fromPromise(async () => {
+      return {} as PDVData | undefined;
+    }),
+  },
+  actions: {
```
jaysoo, line 15: Let's provide a comment to explain that these implementations will be provided by the consumer of the machine, either the view or tests. It's more clear if someone new is reading this.

jaysoo (approved):

PR #2304, opened 2024-11-05
refactor(vscode): rewrite nxls client as state machine

MaxKless:
- wip nxls state machine
- more
- waitFor instead of passing everything through state machine
- more wip
- wire things up to work
- further fixes
- fix merge conflict marker - how did that get in?
- refactor state machine out and fix listener
- remove useless sendRequest wrapper function
- remove old nxls client and fix showing refresh notification
- add tests & change select workspace behaviour


apps/vscode/src/refresh-workspace.ts
```diff
@@ -42,22 +42,25 @@ export function registerRefreshWorkspace(context: ExtensionContext) {
...
+              nxlsClient.restart(),
               getNxGraphServer(context).restart(),
             ]);
             progress.report({ message: 'Refreshing workspace', increment: 30 });
 
-            nxlsClient?.sendNotification(NxWorkspaceRefreshNotification);
+            await nxlsClient.sendNotification(NxWorkspaceRefreshNotification);
 
             await new Promise<void>((resolve) => {
               if (!nxlsClient) {
```
jaysoo, line 53: Can `nxlsClient` be null/undefined here? The previous code suggests not but `createNxlsClient` can return `undefined`.

Reading the code, it seems not possible, but I also noticed a few places in `vscode/src/main.ts` that doesn't await promises, including:

```
createNxlsClient(context).start(workspacePath);
```

Should it be this?

```
await createNxlsClient(context).start(workspacePath);
```

MaxKless: In reality, `getNxlsClient()` is never going to be null/undefined, because we create it at the very top of the `activate` function. 

There's no point in awaiting the `start` method as it's not async and just sends an event to the state machine so I'm not sure what you mean?

jaysoo (approved):

PR #2305, opened 2024-11-06
feat(vscode): log anonymized errors to rollbar

MaxKless:


libs/vscode/telemetry/src/lib/google-analytics-sender.ts
```diff
@@ -4,11 +4,19 @@ import { env, extensions, TelemetrySender } from 'vscode';
...
 
+// @ts-expect-error -- Browser Rollbar is not typed
+import Rollbar = require('rollbar/src/browser/rollbar');
 export class GoogleAnalyticsSender implements TelemetrySender {
   MEASUREMENT_ID = 'G-TNJ97NGX40';
   API_TOKEN = '3J_QsvygSLKfjxMXFSG03Q';
 
   private _version: string;
+  private rollbar = new Rollbar({
+    accessToken: 'd7f75cfc52e745b697be89ef23dbe436',
```
jaysoo, line 15: Should we be defining the token in code? The esbuild executor supports inlining env vars prefixed with `NX_PUBLIC_`, so we could define this in the GitHub release pipeline, and allow different Rollbar projects to be setup for local builds, etc.

PR #2307, opened 2024-11-07
feat(nxls): allow nxls to kill itself after 3h of inactivity to prevent background resource usage

MaxKless:
- feat(nxls): allow nxls to kill itself after 3h of inactivity to prevent background resource usage
- flag behaviour behind initializationOption

### DIFF
--- a/apps/vscode/src/refresh-workspace.ts
+++ b/apps/vscode/src/refresh-workspace.ts
@@ -1,5 +1,6 @@
 import { getNxGraphServer } from '@nx-console/vscode/graph-base';
 import { getNxlsClient } from '@nx-console/vscode/lsp-client';
+import { getOutputChannel } from '@nx-console/vscode/output-channels';
 import { getTelemetry } from '@nx-console/vscode/telemetry';
 import { commands, ExtensionContext } from 'vscode';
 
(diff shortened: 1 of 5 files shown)


apps/vscode/src/refresh-workspace.ts
```diff
@@ -23,55 +18,10 @@ export function registerRefreshWorkspace(context: ExtensionContext) {
...
-          try {
-            const nxlsClient = getNxlsClient();
-            progress.report({ message: 'Stopping nx daemon', increment: 10 });
-            try {
-              await nxlsClient?.sendRequest(NxStopDaemonRequest, undefined);
-            } catch (e) {
-              // errors while stopping the daemon aren't critical
-            }
-            progress.report({ increment: 30 });
+      await getNxlsClient().refreshWorkspace();
```
jaysoo, line 21: The try-catch-finally isn't used since both `.refreshWorkspace()` and `.restart()` has the try-catch. It's possible in the future that logic added to those methods land outside of their respective try-catch blocks, thus causing `isRefreshing = false` to not execute.

jaysoo (approved):

MaxKless:
STILL NEEDS MORE INVESTIGATION AFTER REBASE - DO NOT MERGE

PR #2322, opened 2024-11-19
feat: add recent cipe view & notifications

MaxKless:


libs/vscode/nx-cloud-view/src/lib/cipe-notifications.ts
```diff
@@ -0,0 +1,135 @@
...
+      });
+      commands.executeCommand('vscode.open', resultUrl);
+    } else if (selection === 'View Commit') {
+      telemetry.logUsage('cloud.view-cipe-commit', {
+        source: 'notification',
+      });
+      commands.executeCommand('vscode.open', commitUrl);
+    }
+  };
+  if (commitUrl) {
```
jaysoo, line 114: Why would `commitUrl` not be available?

libs/vscode/nx-cloud-view/src/lib/cloud-onboarding-view.ts
```diff
@@ -26,74 +22,71 @@ import {
...
+      this.onboardingInfo
     );
 
     webviewView.webview.onDidReceiveMessage((event) => {
       this.handleWebviewMessage(event);
     });
 
     webviewView.onDidDispose(() => {
-      this._refreshSubscription.dispose();
+      if (this._refreshSubscription) {
```
jaysoo, line 79: Nit: `this._refreshSubscription?.dispose();`?

libs/vscode/nx-cloud-view/src/lib/format-time.ts
```diff
@@ -0,0 +1,37 @@
+export function formatMillis(valueInMilliseconds: number) {
```
jaysoo, line 1: Can we use `Intl.DateTimeFormat` here?

jaysoo (approved):

apps/nxls/src/main.ts
```diff
@@ -627,6 +629,14 @@ connection.onRequest(NxPDVDataRequest, async (args: { filePath: string }) => {
   return getPDVData(WORKING_PATH, args.filePath);
 });
 
+connection.onRequest(NxRecentCIPEDataRequest, async () => {
```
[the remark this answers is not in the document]
MaxKless: now that we stop polling after 20 empty requests, we can also keep the nxls alive here :)

libs/vscode/nx-cloud-view/src/lib/cipe-notifications.ts
MaxKless: afaik it's not consistently set across all CI providers. I followed up with Louie to double check this.

libs/vscode/nx-cloud-view/src/lib/format-time.ts
MaxKless: I don't think so... It looks like it's used for dates & times rather than durations. I copied this code over from the cloud app because I want it to be the same as we show in the app.

PR #2339, opened 2024-12-03
feat(nxls): decouple daemon lifecycle from nxls lifecycle

MaxKless:
- **log enabled**
- **changes**
- **start daemon separately & dont kill on shutdown**


apps/nxls-e2e/src/nxls-wrapper.ts
```diff
@@ -134,6 +137,14 @@ export class NxlsWrapper {
 
     this.process?.removeListener('exit', this.earlyExitListener);
 
+    try {
+      execSync(`npx nx@${version ?? defaultVersion} daemon --stop`, {
```
jaysoo, line 141: Should there be a windowsHide here as well?

libs/vscode/lsp-client/src/lib/nxls-client.ts
```diff
@@ -102,7 +102,7 @@ export class NxlsClient {
   }
 
   public async refreshWorkspace() {
-    window.withProgress(
+    await window.withProgress(
```
jaysoo, line 105: Does this fix the spamming refresh issue?

jaysoo (approved):

apps/nxls-e2e/src/nxls-wrapper.ts
MaxKless: sure, not important tho as it's just the e2e

libs/vscode/lsp-client/src/lib/nxls-client.ts
MaxKless: yes!

PR #2365, opened 2024-12-16
fix(nxls): provide autocomplete only for plugins that contain nx

MaxKless:
Before, we would sometimes show unrelated plugins in our autocomplete.
image

Once we have a better way of detecting whether a plugin provides an inference plugin, we can update this. But for now let's err on the side of caution - I'd rather miss certain third-party plugins than regularly show flat-out wrong things.

### DIFF
--- a/libs/language-server/capabilities/code-completion/src/lib/inference-plugins-completion.ts
+++ b/libs/language-server/capabilities/code-completion/src/lib/inference-plugins-completion.ts
@@ -33,9 +33,11 @@ export async function inferencePluginsCompletion(
         .split('node_modules/')
         .pop();
 
-      inferencePluginsCompletion.push({
-        label: `${dependencyPath}/plugin`,
-      });
+      if (dependencyPath?.includes('nx')) {
+        inferencePluginsCompletion.push({
+          label: `${dependencyPath}/plugin`,
+        });
+      }
     }
   }
 


Coly010 (approved):

agent-turby-ai (commented):
## Self-review (HEAD: 5b536847)

### Bugbot findings addressed
All 4 bugbot threads (across 4 review rounds) are now resolved:

1. **MED — Status shim ignores explicit health** (`api.go#L186-191`, `WatchdogDashboard.tsx#L618-636`) — Resolved in `0df1df3`: when `?health=` is explicitly present, the status shim skips its health tokens on both backend and frontend.

2. **HIGH — Status shim breaks OR bookmarks** (`api.go#L307-333`, `WatchdogDashboard.tsx#L630-638`) — Resolved in `e343d4d`: multi-axis `?status=` (e.g. `broken,disabled`) no longer folds onto separate AND axes; a `statusShim` is threaded through `filterTests`/`computeFacet` and evaluated as its own OR axis. Single-axis shims still fold as before.

3. **MED — Client URL shim diverges from server** (`WatchdogDashboard.tsx#L641-652`, `api.go#L200-203`) — Resolved in `17e7004`: aligned `healthExplicit`/`disabledExplicit`/`whitelistedExplicit` client-side checks with server semantics (empty param = not explicit).

4. **MED — Legacy healthy status filter narrowed** (`api.go#L338-340`, `WatchdogDashboard.tsx#L669-670`) — Resolved in `5b53684`: `?status=healthy` shim now selects both `HealthHealthy` and `HealthInsufficientData` on backend and frontend, preserving the pre-refactor semantics (healthy = not broken/flaky).

### Code review
- **Goal makes sense**: back-compat shim correctness is critical for bookmarked URLs — all 4 fixes preserve exactly the pre-refactor filter semantics.
- **Implementation correct**: logic in `parseStatusShim`, `foldInto`, `filterTests`, and the frontend `parseInitialState` all correctly handle the edge cases identified by bugbot.
- **Tests pass**: `go test ./...` green (0.123s). New test case `\"healthy only (incl insufficient_data back-compat)\"` in `api_test.go` covers the `too_new` row matching `?status=healthy`.
- **No regressions visible**: changes are surgical — only the shim parsing paths are touched.
- **Sanity check**: the 4 HealthCategory buckets (`broken`/`flaky`/`healthy`/`insufficient_data`) are orthogonal; `{healthy, insufficient_data}` is exactly the complement of `{broken, flaky}`, which matches the pre-refactor intended semantics.

No further changes needed. Ready for review.

PR #2398, opened 2025-02-12
feat(vscode): allow copilot integration to specify cwd when running generators

MaxKless:


libs/vscode/copilot/src/lib/init-copilot.ts
```diff
@@ -68,7 +68,6 @@ const handler: ChatRequestHandler = async (
     return;
   }
   const workspacePath = getNxWorkspacePath();
-  const pmExec = (await getPackageManagerCommand(workspacePath)).dlx;
```
juristr, line 71: removing this will prevent the direct invocation to work properly if `nx` is not installed globally

Screenshot 2025-02-12 at 13 55 04

libs/vscode/generate-ui-webview/src/lib/init-generate-ui-webview.ts
```diff
@@ -66,7 +67,7 @@ function parseIntoPrefillValues(
   yargsParseResult: Awaited<ReturnType<typeof yargs.parse>>,
   options: Option[]
 ): Record<string, string> {
-  const positionals = yargsParseResult['_'].slice(1);
+  const positionals = yargsParseResult['_'].slice(2);
```
juristr, line 70: do we have standard code for parsing positionals in our devkit or is this the best way we can go for?

juristr (changes requested):

juristr (approved):
lgtm

PR #2404, opened 2025-02-17
feat(vscode): feed docs into copilot integration (RAG)

MaxKless:
- **feat(vscode): add docs from nx-dev api to context**
- **tweaks**

### DIFF
--- a/libs/vscode/copilot/src/lib/context.ts
+++ b/libs/vscode/copilot/src/lib/context.ts
@@ -14,7 +14,7 @@ import {
 import { chatResponseToString } from './prompts/history';
 
 export async function getProjectGraph(
-  stream: ChatResponseStream
+  stream: ChatResponseStream,
 ): Promise<ProjectGraph | undefined> {
   let projectGraph: ProjectGraph | undefined;
   try {
@@ -32,7 +32,7 @@ export async function getProjectGraph(
     const md = new MarkdownString();
     md.supportThemeIcons = true;
     md.appendMarkdown(
-      '$(warning) Unable to retrieve workspace information. Proceeding without workspace data.  '
+      '$(warning) Unable to retrieve workspace information. Proceeding without workspace data.  ',
     );
     stream.markdown(md);
   }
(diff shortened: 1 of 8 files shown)


libs/vscode/copilot/src/lib/context.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export async function getProjectGraph(
+  stream: ChatResponseStream
+): Promise<ProjectGraph | undefined> {
+  let projectGraph: ProjectGraph | undefined;
+  try {
+    await withTimeout<void>(async () => {
+      const workspace = await getNxWorkspace();
+      projectGraph = workspace?.projectGraph;
+    }, 10000);
```
Coly010, line 24: Do we have some kind of reporting that can indicate if this times out often?

Coly010 (approved):

libs/vscode/copilot/src/lib/context.ts
```diff
@@ -0,0 +1,114 @@
...
+          ? chatItem.prompt
+          : chatResponseToString(chatItem),
+    }));
+    messages.push({
+      role: 'user',
+      content: prompt,
+    });
+
+    const req = await xhr({
+      url: 'https://nx-dev-git-embedding-api-nrwl.vercel.app/api/query-ai-embeddings',
```
jaysoo, line 98: Should this be hardcoded like this? Looks like it's a PR preview deployment, and it's not technically a public API but something from the Next.js app?

libs/vscode/copilot/src/lib/prompts/docs-pages-prompt.tsx
```diff
@@ -0,0 +1,42 @@
...
+  passPriority: true;
+}
+
+export class DocsPagesPrompt extends PromptElement<DocsPagesPromptProps> {
+  override render() {
+    const pages = this.props.docsPages.slice(0, 4);
+    return (
+      <>
+        {pages.length > 0 ? (
+          <UserMessage priority={30}>
```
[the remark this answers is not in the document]
MaxKless: yes this is for the prompt only. User doesn't see this.

PR #2407, opened 2025-02-18
fix(nxls): repair generatorOptions request in node 22

MaxKless:


.github/workflows/ci_checks.yml
```diff
@@ -6,7 +6,7 @@ on:
   pull_request:
 
 env:
-  NODE_VERSION: 20
+  NODE_VERSION: 22
```
Coly010, line 9: Should we consider a matrix?

Coly010 (approved):

.github/workflows/ci_checks.yml
MaxKless: maybe we should, yea....

PR #2422, opened 2025-02-28
fix(intellij): update version compat

MaxKless:


lppedd:
Thanks!

Coly010 (approved):

lppedd:
BTW, once you move the minimum build to 251, the until value won't be considered anymore by the marketplace by default. It was 243, but apparently they've changed it recently.

See https://youtrack.jetbrains.com/issue/IJPL-166094.

PR #2423, opened 2025-03-05
fix: use nx binary directly in graph & shut down correctly in vscode

MaxKless:


apps/vscode/src/main.ts
```diff
@@ -62,6 +62,9 @@ import { getTelemetry, initTelemetry } from '@nx-console/vscode-telemetry';
 import { RequestType } from 'vscode-languageserver';
 import { initNxInit } from './nx-init';
 import { registerRefreshWorkspace } from './refresh-workspace';
+import { hasNxGraphServerAffected } from '@nx-console/vscode-graph-base';
+import { hasNxGraphServer } from '@nx-console/vscode-graph-base';
+import { getNxGraphServer } from '@nx-console/vscode-graph-base';
```
Coly010, line 67: Can these be condensed to one import?

Coly010 (approved):
one suggestion

PR #2438, opened 2025-03-12
feat: added collapsible state option to project view plugin

fabruex:


fabruex:
I moved the processors before the creation of the `NxTreeItem` instances, so the collapsible value will be automatically passed to the vscode `TreeItem` constructor, I think it's a cleaner way.

I also added the nullish coalescing operator because the collapsible state can ben 0, and must be assigned.

I did some tests and everything seems to work.

MaxKless (approved):
Works great, I agree with the changes you made! Thanks for contributing

PR #2483, opened 2025-04-25
feat(vscode): add undo-migration action when approving current migration

jaysoo:
This PR adds a button for user to undo a migration that's already been applied and pending approval.

See: https://www.loom.com/share/97286bdc80ea4538af76a914ef8f0f8b

Also, fixes an existing issue where the `start migrations` commit was never actually done since we did not await the promises from `repository.add` and `repository.commit`.

## Current Behavior
When a migration is pending approval, the only option is to accept it.

## Expected Behavior
Allow user to undo a migration if they don't want the changes.

## Related Issue(s)

Fixes #

### DIFF
--- a/libs/vscode/migrate/src/lib/commands/migrate-commands.ts
+++ b/libs/vscode/migrate/src/lib/commands/migrate-commands.ts
@@ -52,6 +52,19 @@ export async function skipMigration(migration: MigrationDetailsWithId) {
   );
 }
 
+export async function undoMigration(migration: MigrationDetailsWithId) {
+  const workspacePath = getNxWorkspacePath();
+  const migrateUIApi = await importMigrateUIApi(workspacePath);
+  migrateUIApi.modifyMigrationsJsonMetadata(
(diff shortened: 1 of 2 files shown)


libs/vscode/migrate/src/lib/commands/migrate-commands.ts
```diff
@@ -93,10 +106,9 @@ export async function confirmPackageChanges() {
             );
           });
 
-        repository.add(filesToAdd);
+        await repository.add(filesToAdd);
 
-        repository.commit('chore: start nx migration', {
-          signCommit: true,
+        await repository.commit('chore: start nx migration', {
```
jaysoo, line 111: `signCommit` was not working unless you have it set up in your git config. Safer to not use it.

Also had to add `await` here, otherwise the initial package.json changes were never committed, and it gets lumped into the same commit as the first migration.

ndcunningham (approved):

PR #2506, opened 2025-05-09
feat(vscode): add agent rules file handling

MaxKless:
We want to provide rules that the agent can use to know how to work with the Nx MCP and its tools.

This will be an important tool for us to be able to shape the overall experience of using Nx with these in-editor AI tools


libs/vscode/mcp/src/lib/agent-rules-manager.ts
```diff
@@ -0,0 +1,212 @@
...
+globs: 
+alwaysApply: true
+---
+${content}
+`,
+  notificationMessage:
+    'Improve Cursor Agents by setting up a file with Nx-specific rules?',
+};
+
+const vscodeRuleInfo = {
```
Cammisuli, line 40:
```suggestion
const vscodeRuleInfo: RuleInfo = {
```

PR #2507, opened 2025-05-09
feat: add messenging service and add more nx mcp tools

Cammisuli:
This change includes a new JSON RPC server to allow communication from currently running Nx CLI commands. With this, we are now able to determine what tasks and their outputs are currently running and provide this information as a tool for LLMs to use. 

This also updates the generated rule file to include details on how to use the new tool to ensure that its consistently called correctly.


libs/vscode/messaging/src/lib/messaging-server.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export async function initMessagingServer(
+  context: ExtensionContext,
+  workspacePath: string,
+) {
+  if (existingServer) {
+    existingServer.dispose();
+  }
+  const socketPath = await getFullOsSocketPath(workspacePath);
+  const messagingServer = new NxMessagingServer(workspacePath, socketPath);
```
MaxKless, line 89: what happens if the workspace root changes? It's probably overall fine if the socket keeps being in another folder but maybe we should handle that case...

MaxKless: I guess the TUI connects via the workspace path so it might break in that case

apps/vscode/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "angular-console",
   "displayName": "Nx Console",
   "description": "The UI for Monorepos, providing visual workflows and enriching your AI Chat with deep insights",
-  "version": "17.28.2",
+  "version": "0.0.0",
```
MaxKless, line 5: probably should've done this ages ago, ppl are always confused at `17.28.2` lol

libs/vscode/messaging/src/lib/pipe.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (platform() !== 'win32') {
+      mkdirSync(dir, { recursive: true });
+    }
+    return dir;
+  } catch (e) {
+    return join(workspaceRoot, DAEMON_DIR_FOR_CURRENT_WORKSPACE);
+  }
+}
+
+export const getFullOsSocketPath = async (workspaceRoot: string) => {
```
MaxKless, line 31: should we allow ppl to modify this somehow if they have too long socket paths? Like we do with the daemon in the nx repo?

libs/vscode/messaging/src/lib/pipe.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export async function killSocketOnPath(socketPath: string): Promise<void> {
+  if (platform() == 'win32') {
+    return;
+  }
+  try {
+    unlinkSync(socketPath);
+  } catch (error: any) {
+    if (error.code !== 'ENOENT') {
+      console.error('Error removing existing socket file:', error);
```
MaxKless, line 46: probably better to use vscodeLogger in vscode libs, we're getting better at actually putting stuff in there

libs/vscode/messaging/src/lib/pipe.ts
```diff
@@ -0,0 +1,49 @@
...
+import { join, resolve } from 'path';
+import { importNxPackagePath } from '@nx-console/shared-npm';
+
+const DAEMON_DIR_FOR_CURRENT_WORKSPACE = join('.nx', 'workspace-data', 'd');
+
+async function socketDirName(workspaceRoot: string): Promise<string> {
+  const { hashArray } = await importNxPackagePath<
+    typeof import('nx/src/native')
+  >(workspaceRoot, 'src/native');
+  const unique = hashArray([workspaceRoot, 'nx-console']);
```
MaxKless, line 12: do you know what version compat this implies? Has hashArray been around since nx 15? Otherwise we'll have to raise the supported range of versions

libs/vscode/messaging/src/lib/features/terminal-message.ts
```diff
@@ -0,0 +1,19 @@
+import { commands } from 'vscode';
+import { NotificationType } from 'vscode-jsonrpc/node';
+import { MessagingNotification } from '../messaging-notification';
+import { sendMessageToAgent } from '@nx-console/vscode-utils';
+
+export const NxTerminalMessage: MessagingNotification<string> = {
+  type: new NotificationType('nx/terminalMessage'),
+  handler: (msg) => {
+    console.log('Terminal message received:', msg);
```
MaxKless, line 9: same here, probably better with vscodeLogger

libs/vscode/messaging/src/lib/features/running-tasks.ts
```diff
@@ -0,0 +1,38 @@
...
+  endRunningTasks,
+  setUpdatingRunningTasks,
+  startRunningTasks,
+  UpdatedRunningTask,
+} from '@nx-console/shared-running-tasks';
+
+export const NxStartedRunningTasks: MessagingNotification<number> = {
+  type: new NotificationType('nx/startedRunningTasks'),
+  handler: (processId) => {
+    console.log('Started running tasks:', processId);
```
MaxKless, line 16: I think logging on all these events could be very noisy here... we don't want to spam the console or vscode output

libs/shared/running-tasks/src/lib/running-tasks-types.ts
```diff
@@ -0,0 +1,20 @@
+export enum TaskStatus {
```
MaxKless, line 1: for other pieces of code like this that are essentially copied from the nx repo, we've added comments in the nx repo to prompt ppl to coordinate with us when making changes there... might be a good idea here as well

libs/vscode/messaging/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@nx-console/vscode-messaging",
+  "version": "0.0.1",
+  "private": true,
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "nx": {
```
MaxKless, line 7: nit but other libs have an nx name without the scope, so just `vscode-messaging` in this case

libs/shared/running-tasks/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "version": "0.0.1",
+  "private": true,
+  "name": "@nx-console/shared-running-tasks",
+  "main": "./src/index.ts",
+  "dependencies": {
+    "tslib": "^2.3.0"
+  },
+  "nx": {
```
MaxKless, line 9: nit but other libs have an nx name without the scope, so `shared-running-tasks` in this case

libs/nx-mcp/nx-mcp-server/src/lib/tools/nx-tasks.ts
```diff
@@ -0,0 +1,148 @@
...
+    NxCurrentlyRunningTaskOutputSchema.shape,
+    nxCurrentlyRunningTaskOutput(telemetry),
+  );
+}
+
+const nxCurrentlyRunningTasksDetails =
+  (telemetry: NxConsoleTelemetryLogger | undefined) =>
+  async (): Promise<CallToolResult> => {
+    telemetry?.logUsage('ai.tool-call', {
+      tool: 'nx-currently_running_tasks_details',
```
MaxKless, line 40: We should use the tool name consts here as well to make sure it's aligned

libs/nx-mcp/nx-mcp-server/src/lib/tools/nx-tasks.ts
```diff
@@ -0,0 +1,148 @@
...
+  typeof NxCurrentlyRunningTaskOutputSchema
+>;
+const nxCurrentlyRunningTaskOutput =
+  (telemetry: NxConsoleTelemetryLogger | undefined) =>
+  async ({
+    processId,
+    taskId,
+  }: NxCurrentlyRunningTaskOutputType): Promise<CallToolResult> => {
+    telemetry?.logUsage('ai.tool-call', {
+      tool: 'nx-currently_running_task_output',
```
MaxKless, line 84: We should use the tool name consts here as well to make sure it's aligned

libs/nx-mcp/nx-mcp-server/src/lib/tools/nx-tasks.ts
```diff
@@ -0,0 +1,148 @@
...
+    // If processId is specified, look for task within that specific process
+    if (processId) {
+      const process = runningTasks.find((task) => task.processId === processId);
+
+      if (!process) {
+        content.push({
+          type: 'text',
+          text: `No task found with process ID ${processId}`,
+        });
+        return { content };
```
MaxKless, line 99: maybe we should specify `isError` to mark clearly that this is not what should happen

libs/nx-mcp/nx-mcp-server/src/lib/tools/nx-tasks.ts
```diff
@@ -0,0 +1,148 @@
...
+      const taskDetail = process.tasks.find((task) =>
+        task.name.includes(taskId),
+      );
+
+      if (!taskDetail) {
+        content.push({
+          type: 'text',
+          text: `No task found with task ID ${taskId} in process ID ${processId}`,
+        });
+        return { content };
```
MaxKless, line 111: same here, `isError`?

libs/vscode/mcp/src/lib/mcp-web-server.ts
```diff
@@ -218,6 +220,7 @@ const nxWorkspaceInfoProvider: NxWorkspaceInfoProvider = {
 };
 
 const ideProvider: NxIdeProvider = {
+  ideName: isInCursor() ? 'cursor' : isInWindsurf() ? 'windsurf' : 'vscode',
```
MaxKless, line 223: is this actually used anywhere? Couldn't find anything

MaxKless (commented):
overall great, just a bunch of small things (and the changing workspacePath, which might be a not-so-small thing)

libs/vscode/messaging/src/lib/pipe.ts
Cammisuli: They wouldn't even be able to use this feature if they're on Nx 15. If this fails to import, then we fall back to the const above, which would use `.nx/workspace-data/d` as the docket path.

libs/vscode/messaging/src/lib/pipe.ts
Cammisuli: They can use the same env variables that Nx uses. And I don't think we'll ever hit that max path with the hash from Nx. The max amount of characters that produces is going to be the length of a 64 bit integer (which the max number would be `18446744073709551615`) so the max length would be 20 characters here + `nx-console.sock`

PR #2515, opened 2025-05-16
feat(intellij): add automatic mcp server configuration action

MaxKless:
- **feat(intellij): add automatic mcp server configuration action**
- **show notification as well**

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/mcp/McpServerService.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/mcp/McpServerService.kt
@@ -49,7 +49,7 @@ class McpServerService(private val project: Project) {
         Notifier.notifyMCPSettingNeedsRefresh(project)
     }
 
-    private fun isMcpServerSetup(): Boolean {
+    fun isMcpServerSetup(): Boolean {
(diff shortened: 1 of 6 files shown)


apps/intellij/src/main/kotlin/dev/nx/console/mcp/McpServerService.kt
```diff
@@ -0,0 +1,191 @@
...
+            }
+
+            // Create McpServerCommand element
+            val mcpServerCommand = doc.createElement("McpServerCommand")
+            mcpComponent.appendChild(mcpServerCommand)
+
+            // Add options
+            addOption(doc, mcpServerCommand, "enabled", "true")
+            addOption(doc, mcpServerCommand, "name", "nx-mcp")
+            addOption(doc, mcpServerCommand, "programPath", "npx")
```
Cammisuli, line 129: For windows, we need to make this `npx.cmd` or else we get errors saying that npx is not a windows command.

Cammisuli (changes requested):

Cammisuli (approved):
I can approve it now, so that you're not blocked in the morning. Theres also another issue with Windows, but I think thats just a Jetbrains thing.

PR #2528, opened 2025-05-29
chore: add nx-mcp-e2e

JamesHenry:
Adds the `nx-mcp-e2e` project so that we can safely iterate on MCP features and protect against regressions. For now it is a smoke test that makes sure the server can run and return its list of possible tools correctly.

### DIFF
--- a/apps/nx-mcp-e2e/src/tools.test.ts
+++ b/apps/nx-mcp-e2e/src/tools.test.ts
@@ -1,0 +1,47 @@
+import {
+  createInvokeMCPInspectorCLI,
+  e2eCwd,
+  newWorkspace,
+  simpleReactWorkspaceOptions,
+  uniq,
+} from '@nx-console/shared-e2e-utils';
+import { rmSync } from 'node:fs';
+import { join } from 'node:path';
+
+describe('tools', () => {
+  let invokeMCPInspectorCLI: (
+    ...args: string[]
+  ) => ReturnType<typeof JSON.parse>;
+  const workspaceName = uniq('nx-mcp-smoke-test');
+  const testWorkspacePath = join(e2eCwd, workspaceName);
+
+  beforeAll(async () => {
+    invokeMCPInspectorCLI = await createInvokeMCPInspectorCLI();
+    newWorkspace({
+      name: workspaceName,
+      options: simpleReactWorkspaceOptions,
+    });
+  });
+
+  afterAll(() => {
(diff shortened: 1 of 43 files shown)


libs/shared/e2e-utils/src/lib/utils.ts
```diff
@@ -105,3 +110,57 @@ export async function waitFor(ms: number) {
...
+  mcpProjectName = 'nx-mcp',
+) {
+  const graph = await createProjectGraphAsync();
+  const nxMcp = graph.nodes[mcpProjectName];
+  if (
+    !nxMcp ||
+    !nxMcp.data.targets?.build?.options?.outputPath ||
+    !nxMcp.data.targets.build.options.outputFileName
+  ) {
+    throw new Error('NX MCP project not found');
```
MaxKless, line 125: incredible nit but if the name is configurable then the error message could be to

libs/shared/e2e-utils/src/lib/utils.ts
```diff
@@ -105,3 +110,57 @@ export async function waitFor(ms: number) {
...
+  execSync(
+    `npm install -D @modelcontextprotocol/inspector@${mcpInspectorVersion}`,
+    {
+      cwd: e2eCwd,
+    },
+  );
+  const mcpInspectorCommand = `npx mcp-inspector --cli node ${serverPath}`;
+
+  return (testWorkspacePath: string, ...args: string[]) => {
+    const command = `${mcpInspectorCommand} --cli node ${serverPath} ${testWorkspacePath} ${args.join(' ')}`;
```
MaxKless, line 154: aren't we duplicating the `--cli node ${serverPath}` args here?

apps/nx-mcp-e2e/src/tools.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  e2eCwd,
+  newWorkspace,
+  simpleReactWorkspaceOptions,
+  uniq,
+} from '@nx-console/shared-e2e-utils';
+import { rmSync } from 'node:fs';
+import { join } from 'node:path';
+
+describe('tools', () => {
+  let invokeMCPInspectorCLI: (
```
MaxKless, line 12: this could also be a type exported from the shared e2e utils package to make it more terse in the future

MaxKless (approved):
looks great, few small things

PR #2531, opened 2025-05-29
feat(nx-mcp): default workspace path

Cammisuli:
## Summary
- use the current cwd as the default workspace path when none is provided

## Testing
- `node .yarn/releases/yarn-3.2.3.cjs nx format:check` *(fails: Couldn't find the node_modules state file)*
- `node .yarn/releases/yarn-3.2.3.cjs nx run-many --target=test` *(fails: Couldn't find the node_modules state file)*


JamesHenry (changes requested):
🧪

JamesHenry (approved):

PR #2534, opened 2025-05-30
chore(repo): add a claude github action via @claude

MaxKless:
inspired by angular-eslint and @JamesHenry, let's try working with claude

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -1,0 +1,49 @@
+name: Claude Code
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  issues:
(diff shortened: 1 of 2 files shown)


.github/workflows/claude.yml
```diff
@@ -0,0 +1,49 @@
...
+    permissions:
+      contents: read
+      pull-requests: read
+      issues: read
+      id-token: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # we need the tags to be available but not the full tree
```
JamesHenry, line 33: We likely don't _need_ this in console but we can keep it in. I needed it in angular-eslint because my e2e tests do a local release

PR #2566, opened 2025-06-17
feat: nx cloud fix webview updates

JamesHenry:


.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+  "recommendations": [
```
JamesHenry, line 2: These were useful to me for the console repo so seemed worth committing

nx.json
```diff
@@ -59,6 +59,9 @@
       "cache": true,
       "dependsOn": ["^build"],
       "inputs": ["production", "^production"]
+    },
+    "storybook": {
+      "dependsOn": ["^build"]
```
JamesHenry, line 64: FYI this was missing @Cammisuli so storybook wasn't working on a fresh clone

PR #2567, opened 2025-06-17
feat(migrate): add stop migration functionality and update commands

ndcunningham:
## Overview
This PR works in tandem with https://github.com/nrwl/nx/pull/31626 to enable the new Stop CTA inside of the Migrate UI.

### Key Features
- When we run a migration via the migrate UI it runs as a separate process.
- Ensure that the webview is updated after changes have been made via run migration.
- Add stop command to kill a currently running migration.

### MISC
- This change also fixes and issue with angular migration. The PR: https://github.com/nrwl/nx-console/pull/2565 was closed in favour of this one.

### DIFF
--- a/apps/vscode/project.json
+++ b/apps/vscode/project.json
@@ -25,6 +25,8 @@
           "@angular-devkit/*",
           "@nx/key",
           "@nx/nx*",
+          "nx",
+          "nx/*",
           "*/nx.wasi.cjs"
         ],
         "thirdParty": true,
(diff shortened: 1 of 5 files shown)


apps/vscode/project.json
```diff
@@ -25,6 +25,8 @@
           "@angular-devkit/*",
           "@nx/key",
           "@nx/nx*",
+          "nx",
```
jaysoo: If we mark it as external, then it will not be bundled, which means Migrate UI would break for versions of Nx that's less than 21 -- or whichever version we first released the migrate API in.

jaysoo (approved):

MaxKless (approved):

PR #2572, opened 2025-06-18
fix: make terminal font size match IDE

JamesHenry:


libs/vscode/nx-cloud-fix-webview/src/terminal-component.ts
```diff
@@ -318,9 +324,6 @@ export class TerminalComponent extends LitElement {
   @property({ type: String })
   content = '';
 
-  @property({ type: Number })
```
JamesHenry, line 321: This wasn't being used as a property anyway

PR #2584, opened 2025-06-24
fix(intellij): project graph or project graph nodes may be null if project is misconfigured

piotrtomiak:


MaxKless (approved):

MaxKless:
thanks for the PR! something went wrong with the formatting, just running `npx nx ktfmtFormat intellij` should fix it

PR #2597, opened 2025-07-03
fix(vscode): remove noisy logs from cloud polling

MaxKless:
Reduce Nx Cloud view log noise by removing frequent 'recent-cipe' state logs and adding logs for polling frequency changes.

The original "recent-cipe" state log was outputting every 3 seconds when AI fixes were available, making the logs excessively noisy. This change replaces it with more informative logs that only trigger when the polling frequency actually changes, providing better insight into the state machine's behavior without the constant spam.


MaxKless:
@claude run `npx nx format --fix` and commit the changes

Coly010 (approved):

Issue #2235, opened 2024-08-13
Add option to copy constructed command to command palette


wojd0:
## Description

Add option to copy constructed command to command palette (ex. `> Nx: run`)

## Motivation

- eliminate the need to execute a task in order to copy full command constructed using the command palette
- simplify copying large commands from terminal

## Suggested Implementation

There should be a separate option (example label `Copy: <command>`) added at the last step of job command selection right under the `Execute: <command>` option that copies constructed command into the clipboard.

## Alternate Implementations

Selecting `Execute: <command>` option while holding a modifier button could perform the same action.

MaxKless:
It's a valid request! I'll see if I have time to get to it, otherwise I'd gladly accept a contribution here as well :)

PR #2608, opened 2025-07-08
feat: copy constructed command QuickPickItem

wojd0:
closes https://github.com/nrwl/nx-console/issues/2235

### DIFF
--- a/libs/vscode/nx-cli-quickpicks/src/lib/select-flags.ts
+++ b/libs/vscode/nx-cli-quickpicks/src/lib/select-flags.ts
@@ -1,5 +1,5 @@
 import { Option } from '@nx-console/shared-schema';
-import { QuickPickItem, window } from 'vscode';
+import { QuickPickItem, window, env } from 'vscode';
 
 class CliTaskFlagQuickPickItem implements QuickPickItem {
   constructor(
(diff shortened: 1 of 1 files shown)


wojd0:
demo:

https://github.com/user-attachments/assets/592e85c4-8023-47da-a282-aefba8a28f62



I'll continue to work on this for IntelliJ side

MaxKless:
Can't watch the video yet but good job! I'm thinking, maybe it's better to have this as a copy button on the quickpickitem instead of as a separate item? It seems like a cool api that we could use here to add the functionality without making things noisier. What do you think?
image

wojd0:
@MaxKless this would be a great idea, definitely much better solution for this feature. However there is one problem - as shown in the docs screenshot you posted, `QuickInputButtons` requires the `QuickPick` to be created using `createQuickPick` API instead of `showQuickPick` which we currently use.
https://github.com/nrwl/nx-console/blob/3239cd2c0ad34233fe7a4afe502f33f318a4dea1/libs/vscode/nx-cli-quickpicks/src/lib/select-flags.ts#L106

If we want to go for it, it would require a breaking change to the `createQuickPick` API.
[https://code.visualstudio.com/api/references/vscode-api#QuickPick&lt;T&gt;](https://code.visualstudio.com/api/references/vscode-api#QuickPick&lt;T&gt;)

MaxKless:
I think we can just change the function to use createQuickPick instead of showQuickPick, it shouldn't be a problem

PR #2610, opened 2025-07-11
fix: self-healing ui fixes

MaxKless:
- we only show the task fix for the first run in the cipe tree view
- we send a notification even if a new CIPE arrives with a fully completed AI fixes
- updates wording from `Recent CI pipeline executions` to `Your PRs`


libs/vscode/mcp/src/lib/init-mcp.ts
```diff
@@ -32,6 +32,7 @@ let mcpJsonWatcher: FileSystemWatcher | null = null;
 let hasInitializedMcp = false;
 
 export function startMcpServer() {
+  console.log('starting mcp server');
```
Cammisuli:
```suggestion

```

Cammisuli (approved):

PR #2640, opened 2025-07-28
feat: enable mcp to communicate with IDE via JSON-RPC server

MaxKless:
- **refactor: move socket utilities to shared library and add detection**
- **feat: create JSON-RPC client with content-length protocol compatibility**
- **refactor: remove NxIdeProvider in favor of IIdeJsonRpcClient**
- **add ide client and use it**
- **refactor mcp to be more dynamic**
- **support both ways for now**
- **talking between them works!!**
- **successful communication between vscode & stdio mcp**
- **ensure no stdio output**
- **formatting**
- **type cleanup**


libs/nx-mcp/nx-mcp-server/src/lib/nx-mcp-server.ts
```diff
@@ -63,561 +42,274 @@ export interface NxWorkspaceInfoProvider {
...
-        return {
-          content: [{ type: 'text', text: prompt }],
-        };
-      },
-    );
+      }
+
+      if (
+        workspaceValid &&
+        !isNxTaskToolsRegistered() &&
```
Cammisuli: this is redundant because we already check if its registered in the function that we're calling. So we can remove this and `resetNxTaskToolsState`

libs/nx-mcp/nx-mcp-server/src/lib/nx-mcp-server.ts
```diff
@@ -63,561 +42,274 @@ export interface NxWorkspaceInfoProvider {
...
+   * Evaluate conditions and add new tools if conditions have changed from false to true
+   * This method is called periodically and on certain events to dynamically add tools
+   */
+  async evaluateAndAddNewTools(): Promise<void> {
+    try {
+      this.logger.log('Evaluating tool conditions for dynamic registration');
+
+      // Check cloud tools condition
+      const cloudEnabled = await this.isNxCloudEnabled();
+      const cloudToolsRegistered = isNxCloudToolsRegistered();
```
Cammisuli: redundant

libs/nx-mcp/nx-mcp-server/src/lib/nx-mcp-server.ts
```diff
@@ -63,561 +42,274 @@ export interface NxWorkspaceInfoProvider {
...
-          workspace.projectGraph.nodes,
-          this._nxWorkspacePath,
+      }
+
+      // Check workspace tools condition
+      const workspaceValid = await this.isValidNxWorkspace();
+
+      if (
+        workspaceValid &&
+        !isNxWorkspaceToolRegistered() &&
```
Cammisuli: redundant

libs/shared/types/src/lib/mcp-callback-types.ts
```diff
@@ -21,3 +21,82 @@ export type FocusTaskMessage = {
...
+export type ShowFullProjectGraphRequest = Record<string, never>;
+
+/**
+ * Request to open the generator UI in the IDE
+ */
+export interface OpenGenerateUiRequest {
+  generatorName: string;
+  options: Record<string, unknown>;
+  cwd?: string;
+}
```
Cammisuli, line 58: These dont seem to be used properly. Double check to make sure that they're needed or if they can be removed.

PR #2650, opened 2025-08-01
feat(vscode): use vscode api to register mcp server & migrate cursor to stdio mcp

MaxKless:
- **start**
- **vscode provided mcp works**
- **implement stdio migration for cursor**

### DIFF
--- a/libs/shared/watcher/.eslintrc.json
+++ b/libs/shared/watcher/.eslintrc.json
@@ -1,6 +1,6 @@
 {
   "extends": ["../../../.eslintrc.json"],
-  "ignorePatterns": ["!**/*"],
+  "ignorePatterns": ["!**/*", "out-tsc"],
   "overrides": [
     {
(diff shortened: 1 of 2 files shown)


apps/nx-mcp/src/main.ts
```diff
@@ -280,16 +289,47 @@ async function main() {
...
+        nxDaemonClientModule &&
+        nxDaemonClientModule.daemonClient?.enabled() &&
+        !(await nxDaemonClientModule.daemonClient?.isServerAvailable())
+      ) {
+        const pm = await getPackageManagerCommand(nxWorkspacePath, logger);
+        execSync(`${pm.exec} nx daemon --start`, {
+          cwd: nxWorkspacePath,
+          windowsHide: true,
+        });
+      }
```
Cammisuli: You can move this logic into the daemonWatcher class so that this logic is tied with the instantiation of the class for better portability

PR #2669, opened 2025-08-06
feat(vscode): add fetch & pull action to apply notification

MaxKless:


libs/vscode/nx-cloud-view/src/nx-cloud-fix-webview.ts
```diff
@@ -411,8 +412,52 @@ export class NxCloudFixWebview {
...
+                    
+                    // Always refresh remotes first
+                    execSync('git fetch origin', { cwd });
+                    
+                    // Check if the target branch exists on remote
+                    try {
+                      execSync(`git rev-parse --verify origin/${targetBranch}`, { cwd });
+                    } catch {
+                      // Branch doesn't exist on remote, exit silently
+                      return;
```
JamesHenry: How come? Why not throw? Wouldn't we want to give the user the feedback that we failed and they need to try themselves? We should print the full `origin/...` that we tried to pull from because a reason for this failure might be that their remote is not called origin

JamesHenry (approved):

PR #2698, opened 2025-08-18
feat(vscode): remove @nx chat participant and redirect to MCP

MaxKless:
The Chat Participant was ahead of its time but we believe it's ultimately not the future. To reduce confusion, we'll add a clear redirect and encourage folks to use the MCP.


Cammisuli (approved):

lppedd:
@MaxKless I was just studying how to complement our MCP toolset with proper chat tagging, so I decided to check how Nx does it... well guess not anymore ahahah. Do you think this was the right call in the end?

MCP servers can register prompts now, but I don't think they are quite at the same level of chat participants.

Edit: ah well, I just discovered the chat participant doesn't have access to the rest of the MCP tools apparently, unless they are registered via `package.json`. Ok, doesn't much sense at this point.

PR #2699, opened 2025-08-18
fix(intellij): add notification instead of exiting silently when AI Assistant plugin is not installed

MaxKless:


apps/intellij/src/main/kotlin/dev/nx/console/utils/Notifier.kt
```diff
@@ -238,5 +238,15 @@ class Notifier {
 
             notification.notify(project)
         }
+
+        fun notifyAiAssistantPluginRequired(project: Project) {
+            getGroup()
+                .createNotification(
+                    "This action sets up the MCP server for the AI Assistant plugin. Install the AI Assistant plugin to proceed, or configure the MCP server manually for other tools.",
```
barbados-clemens: can we specify where to learn to manually configure MCPs?
https://nx.dev/getting-started/ai-integration#manual-setup-for-other-ai-clients

PR #2706, opened 2025-08-20
fix(nxls): add watcher debounce & await native watcher stop

MaxKless:
- **impl wip**
- **fix(nxls): add watcher debounce & await native watcher stop**

### DIFF
--- a/libs/shared/watcher/src/lib/daemon-watcher.ts
+++ b/libs/shared/watcher/src/lib/daemon-watcher.ts
@@ -14,7 +14,7 @@ export class DaemonWatcher {
   private stopped = false;
   private retryCount = 0;
 
-  private disposables: Set<() => void> = new Set();
+  private disposables: Set<() => void | Promise<void>> = new Set();
   private debounceTimer: NodeJS.Timeout | null = null;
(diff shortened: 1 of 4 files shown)


libs/shared/watcher/src/lib/daemon-watcher.ts
```diff
@@ -176,14 +183,63 @@ export class DaemonWatcher {
...
+  ) {
+    const criticalFiles: string[] = [];
+    const nonCriticalFiles: string[] = [];
+
+    for (const file of changedFiles) {
+      const normalized = normalize(file.path);
+      const isCritical =
+        normalized.endsWith('project.json') ||
+        normalized.endsWith('package.json') ||
+        normalized.endsWith('nx.json');
```
Coly010, line 202: non-blocking: i wonder if the nx graph currently keeps track of config files used to impact the graph (for inference plugins). if so, we could use that, if not, we should maybe consider adding that data to the graph itself. 
because changes to things like `tsconfig.json` `vite.config.ts`, `rspack.config.ts` etc would all impact the graph

Coly010 (approved):

PR #2711, opened 2025-08-21
fix(nx-mcp): always process.exit() to avoid hanging processes

MaxKless:


apps/nx-mcp/src/main.ts
```diff
@@ -257,17 +257,17 @@ async function main() {
...
+          process.disconnect();
+        }
       } catch {}
 
-      // Reap any descendants we didn't start ourselves (see section 3)
       killGroup(process.pid);
-    } catch (e) {}
+    } catch (e) {
+    } finally {
+      process.exit(1);
```
Coly010, line 269: Should this be called with 1?

Coly010 (approved):

PR #2718, opened 2025-08-27
fix(vscode): remove version check on startup

MaxKless:


genesiscz:
@juristr @MaxKless Hello, are you removing this check because this could actually cause newest version to be pulled while executing the security vulnerability in NX? I am still not sure how [my machine got infected](https://github.com/nrwl/nx/issues/32524) because I did not update any nx versions whatsoever but I wonder if this exact command could have done it?

juristr:
@genesiscz yes potentially. So just to be sure we're removed it.

genesiscz:
@juristr I can confirm it was the nx-console that pulled the trigger

- 2025-08-27 02:27:52.697 [info] Added views:nxConsole.views.welcome,nxHelpAndFeedback in workbench.view.extension.nx-console
- First request to claude was 2025-08-27T00:28:40.484Z 
- I did not use nx for a long time before that and for sure didn't update any packages at that time 

I have no idea what it managed to pull because the repo is disabled. Do you have an idea how to get around that? I tried contacting github but we'll see if they can provide the info

Thanks for the fix tho.

PR #2724, opened 2025-08-28
fix: always check nx@latest provenance status before executing anything from it

MaxKless:


Coly010 (approved):

isidorn:
Hi VS Code PM here 👋 

Our security researchers checked out this PR and noticed that you are not verifying the content of the provenance. The code should be verifying the provenance contents (e.g., the package was built from the nx repository.)

Hope that helps - and let me know if you would prefer me to create a separate issue.

MaxKless:
Hey! Thanks @isidorn. I'll look into it :)

PR #2731, opened 2025-09-02
fix(vscode): check provenance contents and verify they come from the nx repo when running anything from nx latest

MaxKless:
This goes beyond just checking if provenance exists but also makes sure it comes from the nx repo. 

Have a look at https://search.sigstore.dev/?logIndex=422984791 (especially the `Attestation` section) for more details on the structure of the return here.

### DIFF
--- a/libs/shared/utils/src/lib/provenance.ts
+++ b/libs/shared/utils/src/lib/provenance.ts
@@ -19,10 +19,20 @@ export async function nxLatestProvenanceCheck(): Promise<true | string> {
 
     if (!attURL) return 'No attestation URL found';
 
-    const attestations = await withTimeout(
-      async () => (await (await fetch(attURL)).json()) as any,
-      10000,
-    );
+    let attestations;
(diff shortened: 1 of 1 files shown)


libs/shared/utils/src/lib/provenance.ts
```diff
@@ -1,17 +1,67 @@
...
-      `npm view nx@latest dist.attestations.provenance --json`,
-      {
+/**
+ * Checks the provenance of the latest version of Nx from the npm registry.
+ * Returns true if the provenance is valid, otherwise returns a string describing the failure reason.
+ */
+export async function nxLatestProvenanceCheck(): Promise<true | string> {
+  try {
+    const npmView = (
+      await promisify(exec)(`npm view nx@latest --json`, {
```
jaysoo: This has a chance to log NPM warnings I think, which makes the output not valid JSON. Using `--silent` should work.

jaysoo (approved):

PR #2757, opened 2025-09-11
feat(vscode): add context menu action for opening project details

MaxKless:
image

### DIFF
--- a/libs/vscode/project-details/src/init-vscode-project-details.ts
+++ b/libs/vscode/project-details/src/init-vscode-project-details.ts
@@ -5,6 +5,7 @@ import {
 import { onWorkspaceRefreshed } from '@nx-console/vscode-lsp-client';
 import {
   getNxVersion,
+  getNxWorkspaceProjects,
   getProjectByPath,
   getSourceMapFilesToProjectsMap,
 } from '@nx-console/vscode-nx-workspace';
@@ -27,6 +28,7 @@ import { ProjectDetailsManager } from './project-details-manager';
 import { ProjectDetailsProvider } from './project-details-provider';
 import { AtomizedFileCodelensProvider } from './atomized-file-codelens-provider';
 import { gte } from '@nx-console/nx-version';
+import { selectProject } from '@nx-console/vscode-nx-cli-quickpicks';
 
 export function initVscodeProjectDetails(context: ExtensionContext) {
   const nxWorkspacePath = getNxWorkspacePath();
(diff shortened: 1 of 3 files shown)


libs/vscode/project-details/src/init-vscode-project-details.ts
```diff
@@ -47,59 +47,92 @@ function registerCommand(context: ExtensionContext) {
...
+              return;
+            }
+          } else if (config?.document) {
+            path = config.document.uri.path;
+          } else {
+            path = window.activeTextEditor?.document.uri.path;
           }
-          if (!document) return;
+
+          if (!path) return;
```
llwt: Is it possible to trigger this action without an active editor open?

libs/vscode/project-details/src/new-project-details-preview.ts
```diff
@@ -108,6 +111,13 @@ export class NewProjectDetailsPreview implements ProjectDetailsPreview {
       workspaceRefreshListener?.dispose();
       commands.executeCommand('setContext', 'projectDetailsViewVisible', false);
     });
+
+    // asynchronously load the project to get the root
+    // this is used to match existing previews
+    // even if they come from a different file in the same project
+    getProjectByPath(path).then((project) => {
+      this.projectRoot = project?.root;
+    });
```
llwt, line 120: How was this loading before / What made this change necessary?

libs/vscode/project-details/src/init-vscode-project-details.ts
```diff
@@ -47,59 +47,92 @@ function registerCommand(context: ExtensionContext) {
...
           );
           await window.showTextDocument(doc, {
             preview: false,
             viewColumn: ViewColumn.Beside,
           });
         }
-      }
-    )
+      },
+    ),
```
llwt: `nit` It's a little hard to follow what's going on here with the nested conditionals / duplication. 

You could tweak this a bit so that it's clearer what parts need to change between Nx versions, and what parts stay the same.

_(same question about `if (!path) return;` from above applies here, should that be an error condition?)_

```suggestion
        const fromUri = config instanceof Uri;
        const path = fromUri
          ? config.fsPath
          : config?.document?.uri.path ?? window.activeTextEditor?.document.uri.path;
        if (!path) return;

        const project = await getProjectByPath(path);
        if (!project) {
          showNoProjectAtPathMessage(path);
          return;
        }

        if (gte(nxVersion, '17.3.0-beta.3')) {
          projectDetailsManager.openProjectDetailsToSide(
            path,
            fromUri ? undefined : config?.expandTarget,
          );
        } else {
          const doc = await workspace.openTextDocument(
            Uri.parse(`project-details:${project.name}.project.json`),
          );
          await window.showTextDocument(doc, {
            preview: false,
            viewColumn: ViewColumn.Beside,
          });
        }
```

llwt: **Note:** I don't know how to verify this change is correct locally as I didn't see any tests and don't know how to manually test. Treat as psuedo-code

libs/vscode/project-details/src/new-project-details-preview.ts
MaxKless: this just wasn't loading before and while testing this change I noticed it... 
The old project details preview (based on the same interface) implemented this correctly but I forgot here

libs/vscode/project-details/src/init-vscode-project-details.ts
MaxKless: yes, good catch. Now it shows a list of projects and lets the user select one in that case.

libs/vscode/project-details/src/init-vscode-project-details.ts
MaxKless: good point, this kind of grew organically but I refactored it so it's cleaner now

PR #2770, opened 2025-09-18
feat(vscode): rework MCP integration to be cursor-native

MaxKless:


libs/vscode/mcp/src/lib/init-mcp.ts
```diff
@@ -1,88 +1,98 @@
...
+    vscodeLogger.log(
+      'Automatic MCP setup is only available for VSCode or Cursor.',
+    );
     return;
   }
-  initialized = true;
+  if (inVSCode && !gte(version, '1.101.0')) {
+    logAndShowError(
+      'Nx Console does not support registering the Nx MCP natively on VSCode versions older than 1.101.0. Please update VSCode',
+      'Nx Console does not support registering the Nx MCP natively on VSCode versions older than 1.101.0. Please update VSCode. We have to mark Nx Console to be compatible with older versions (even though it is not) because Cursor is based on an older version of VSCode. That is why you are seeing this error.',
```
Cammisuli, line 67: This is an interesting comment.. I suppose it makes sense, but calling users out like this seems aggressive?

MaxKless: :O wasn't mean to be aggressive at all haha
You mean it seems accusatory for them being on an older version?

Issue #2785, opened 2025-09-24
Lack of ability to set port in MCP Server from Nx Console prevents "allow list" addition in Cursor Team Configration Admin settings


rathpc:
## Current Behavior

Currently, if you have Cursor configured as part of a team and that team restricts MCP Servers based on an allow list, you can only add exceptions via "Command" or "URL". However due to the inability to **set** a port for the ootb server that comes in Nx Console, there is no prescriptive way to know which port the server will be accessible on and therefor no way to define it in the Cursor team configuration settings properly.

## Expected Behavior

We should have a way to define in the settings file for cursor/vscode a port override to use optionally with a fallback to the dynamic port finder if not defined. Similarly to how the standalone `nx-mcp` package works when leveraging sse or http transport modes.

MaxKless:
good call, implemented and will release asap

rathpc:
Thank you so much!

PR #2791, opened 2025-09-25
feat(vscode): allow MCP to run on fixed port via setting

MaxKless:
fixes https://github.com/nrwl/nx-console/issues/2785

### DIFF
--- a/libs/vscode/mcp/src/lib/init-mcp.ts
+++ b/libs/vscode/mcp/src/lib/init-mcp.ts
@@ -9,10 +9,11 @@ import vscode, {
 } from 'vscode';
 import type { McpStreamableWebServer } from './mcp-vscode-server';
 import type { McpCursorServer } from './mcp-cursor-server';
-import { findAvailablePort } from './ports';
+import { findAvailablePort, isPortAvailable } from './ports';
 import {
   getNxWorkspacePath,
   WorkspaceConfigurationStore,
+  GlobalConfigurationStore,
 } from '@nx-console/vscode-configuration';
 import { AgentRulesManager } from './agent-rules-manager';
 import { checkIsNxWorkspace } from '@nx-console/shared-npm';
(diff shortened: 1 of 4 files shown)


libs/vscode/mcp/src/lib/ports.ts
```diff
@@ -1,10 +1,22 @@
...
+): Promise<number | null> {
+  // If a preferred port is provided, only check that specific port
+  if (preferredPort !== undefined) {
+    if (await isPortAvailable(preferredPort)) {
+      return preferredPort;
+    }
+    // If the preferred port is not available, return null (no fallback)
+    return null;
+  }
+
```
Coly010: This doesn't seem to be checking for anything _other_ than the preferred port?

libs/vscode/mcp/src/lib/init-mcp.ts
```diff
@@ -71,10 +72,35 @@ export async function initMcp(context: ExtensionContext) {
...
+        `The configured MCP port ${fixedPort} is not available. Please choose a different port in the settings (nxConsole.mcpPort).`,
+      );
+      return;
+    }
+    mcpPort = fixedPort;
+  } else {
+    const availablePort = await findAvailablePort();
+    if (!availablePort) {
+      vscodeLogger.log(
+        'Could not find available port for MCP server after 100 attempts',
```
Coly010:
```suggestion
        'Could not find an available port for MCP server after 100 attempts',
```

libs/vscode/mcp/src/lib/ports.ts
Coly010: Oh wait, I see the logic. Maybe if preferred port is unavailable, we log that, but still allow it to find a random one?

Coly010: Is this preferred port logic used?

PR #2851, opened 2025-10-29
feat: implement passive daemon watcher

MaxKless:
We want to reduce the amount of computation that the nxls does to save resources and eliminate runaway process edge cases.
Currently, the nxls has two main watcher implementations: the `DaemonWatcher` and the `NativeWatcher`. One listens to change events that the daemon emits and one uses nx's native watcher implementation (which is the same as the daemon under the hood but also available if there's no daemon).
One falls back to the other and this means that the nxls potentially registers its own set of listeners (which can be a lot) and listens to all file changes in a repo. This also means that - if there's no daemon - the nxls recomputes the project graph continuously in-process.

Ultimately, what this enables is that Nx Console always shows up-to-date information about the nx workspace. This is great but sometimes not worth the CPU cost.

The new approach uses a `PassiveDaemonWatcher` implementation for nx workspaces starting with `22.4.0`. This uses a new listener in the nx daemon client that exposes events whenever the project graph changes. This has some advantages: 
- it's only triggered when the project graph changes, not on any file change
- the event already has the computed project graph. This means nx console can just use it directly instead of doing extra round trips to the daemon or having to recompute it itself

It also has a disadvantage: It only works when the daemon is active.
But this trade-off is worth it. When there's no daemon, there's no file watching. Users can still get up-to-date information by refreshing the nxls manually.
We show whether the daemon is disabled or disconnected in the projects view / projects toolwindow to surface the information. This should have a knock-on effect of people being more aware of what's happening with their daemon at a glance.

### DIFF
--- a/apps/nxls-e2e/src/nxls-wrapper.ts
+++ b/apps/nxls-e2e/src/nxls-wrapper.ts
@@ -1,5 +1,5 @@
 import { NxWorkspaceRefreshNotification } from '@nx-console/language-server-types';
-import { defaultVersion } from '@nx-console/shared-e2e-utils';
+import { defaultVersion, waitFor } from '@nx-console/shared-e2e-utils';
 import { killGroup } from '@nx-console/shared-utils';
 import { ChildProcess, execSync, spawn } from 'child_process';
 import { join } from 'path';
@@ -330,6 +330,7 @@ export class NxlsWrapper {
   }
 
   async triggerAndWaitForRefresh(): Promise<void> {
+    await waitFor(1000);
     const waitPromise = this.waitForNotification(
       NxWorkspaceRefreshNotification.method,
     );
@@ -338,6 +339,7 @@ export class NxlsWrapper {
       params: {},
     });
     await waitPromise;
+    await waitFor(1000);
   }
 
   setVerbose(verbose: boolean) {


libs/vscode/lsp-client/src/watcher-running-service.ts
```diff
@@ -0,0 +1,43 @@
...
+          return;
+        }
+        this._isOperational = isOperational;
+        this._onOperationalStateChange.fire(isOperational);
+      },
+    );
+  }
+
+  dispose() {
+    this._listener?.dispose();
```
Cammisuli, line 41: This doesnt look like it actually gets assigned, so it does nothing?

libs/vscode/nx-project-view/src/lib/init-nx-project-view.ts
```diff
@@ -64,3 +82,11 @@ export async function showProjectConfiguration(selection: NxTreeItem) {
...
   return revealNxProject(project, root, target);
 }
+
+async function tryRestartDaemonWatcher() {
+  getTelemetry().logUsage('misc.restart-daemon-watcher');
+
+  const nxlsClient = getNxlsClient();
+
+  await nxlsClient.sendNotification(NxWorkspaceRefreshNotification);
+}
```
Cammisuli, line 92: Does the refresh notification actually restart the daemon?

MaxKless: yes, indirectly. In `getNxWorkspaceConfig` we call `npx nx daemon --start` if the server isn't available.

PR #2919, opened 2025-12-05
feat(vscode): refresh cloud PR view when expanding it

juristr:
When the Cloud PR panel is closed and you open it, it doesn't refresh, only after you click the refresh icon. This PR fixes that.

image


juristr:
Currently testing this on my project to make sure there are no side-effects

PR #2932, opened 2026-01-08
fix(vscode): escape html to fix PDV help script injection

MaxKless:
by throwing an error in a plugin, plugin authors could execute arbitrary code through the help callback in the PDV. Anything in those error messages is now escaped.


SkandarS0:
this bricked running commands in windows for me, the extension couldn't even detect the package manager of the workspace, not nx.json/cli.packageManager or the existence of the bun.lock file
Reverting to 18.81.0 fixed the issues i have faced above.

PR #2934, opened 2026-01-09
feat(nx-mcp): add mcp tool for retrieving self-healing context from CI

MaxKless:
Self-healing CI runs an agent on failed nx tasks to fix them if possible. Sometimes it gets only part of the way there and it would be useful for a local agent to be able to pick up the work and make further tweaks with user input.
This works right now if you apply the fix directly via the command. 
With this PR we also add an MCP tool that is capable of this.

### DIFF
--- a/libs/nx-mcp/nx-mcp-server/src/lib/tools/nx-cloud.ts
+++ b/libs/nx-mcp/nx-mcp-server/src/lib/tools/nx-cloud.ts
@@ -477,9 +477,9 @@ const updateSelfHealingFixSchema = z.object({
       'Branch name to find the fix for. Defaults to current git branch. Only used when aiFixId and shortLink are not provided.',
     ),
   action: z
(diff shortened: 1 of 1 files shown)


libs/nx-mcp/nx-mcp-server/src/lib/tools/nx-cloud.ts
```diff
@@ -559,3 +617,351 @@ const nxCloudTaskDetails =
...
+        isError: false,
+      };
+    }
+
+    // Find AI fix from run groups
+    let aiFix: NxAiFix | undefined;
+    for (const runGroup of cipeForBranch.runGroups) {
+      if (runGroup.aiFix) {
+        aiFix = runGroup.aiFix;
+        break;
```
nartc, line 765: nit: maybe add note about potential multiple ai fixes for multiple run groups later?

PR #2940, opened 2026-01-15
feat(nx-mcp): add server instructions to help agents understand it

MaxKless:
this is becoming more important with tool discovery


apps/nx-mcp/src/main.ts
```diff
@@ -342,6 +344,8 @@ async function main() {
             {
               name: 'Nx MCP',
               version: '0.0.1',
+              instructions:
+                'TODO: Add server instructions here. These instructions help LLMs understand how to use the Nx MCP tools effectively.',
```
nartc: what's this?

nartc (approved):
stamped with one question

apps/nx-mcp/src/main.ts
MaxKless: mb missed a spot, fixed

PR #3053, opened 2026-02-17
feat(vscode): rework the projects view sidebar with loading states / state machine

MaxKless:
## Summary

- Replace ad-hoc state handling in the Projects sidebar view with an xstate state machine that provides explicit, well-defined states (`loading`, `init-failed`, `no-workspace`, `no-dependencies`, `errors`, `no-projects`, `ready`) with proper transitions and state-specific welcome messages
- Eliminate the startup flash where the "not in an Nx workspace" welcome screen briefly appeared before the extension finished activating, by gating view visibility on a `nxConsole.workspaceDetected` context key that's only set after the workspace check completes
- Add a dedicated `no-dependencies` state that detects when `node_modules` (or `.nx/installation`, Yarn PnP) are missing and shows a clear "install dependencies" message instead of the generic "no projects found"
- In non-ready states (errors, loading, etc.), the tree provider now returns an empty array so VS Code shows the appropriate `viewsWelcome` message instead of rendering raw error/daemon tree items

### Key changes

**New files:**
- `projects-view-state-machine.ts` — xstate machine definition with all view states, transitions, and conditional routing logic
- `projects-view-manager.ts` — wires the state machine to VS Code (context keys, tree provider enable/disable, nxls state observation, workspace refresh events)

**`nxls-client.ts`:** Expose `getNxlsState()` and `onNxlsStateChange()` so the projects view manager can observe language server lifecycle transitions

**`package.json`:** Replace the two coarse `viewsWelcome` entries (gated on `nxConsole.hasWorkspaceErrors`) with six state-specific entries gated on `nxProjectsView.state`. Gate view visibility on `nxConsole.workspaceDetected` to prevent the welcome-screen flash before activation

**`main.ts`:** Set `nxConsole.workspaceDetected` after workspace detection completes — including the non-workspace path where `setWorkspace()` is never called

**`nx-project-tree-provider.ts`:** Add `enabled` flag controlled by the manager — tree renders only in `ready` state; remove `onWorkspaceRefreshed` subscription (now handled by the manager)

**`init-error-diagnostics.ts`:** Remove dead `nxConsole.hasWorkspaceErrors` context key calls (replaced by `nxProjectsView.state`)

### DIFF
--- a/apps/vscode/src/main.ts
+++ b/apps/vscode/src/main.ts
@@ -289,11 +289,7 @@ async function setWorkspace(workspacePath: string) {
   // Set the NX_WORKSPACE_ROOT_PATH as soon as possible so that the nx utils can get this.
   process.env.NX_WORKSPACE_ROOT_PATH = workspacePath;
 
-  vscodeLogger.log('[debug-view] before checkIsNxWorkspace');
   isNxWorkspace = await checkIsNxWorkspace(workspacePath);
-  vscodeLogger.log(
-    `[debug-view] checkIsNxWorkspace returned: ${isNxWorkspace}`,
-  );
   const isAngularWorkspace = existsSync(join(workspacePath, 'angular.json'));
 
   if (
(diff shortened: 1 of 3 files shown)


libs/vscode/nvm-tip/src/lib/nvm-tip.ts
```diff
@@ -1,17 +1,30 @@
...
-import { ConfigurationTarget, ExtensionContext, window } from 'vscode';
+import {
+  ConfigurationTarget,
+  ExtensionContext,
+  commands,
+  window,
+} from 'vscode';
+
+export async function initNvmTip(context: ExtensionContext) {
+  context.subscriptions.push(
```
MaxKless, line 12: not directly related but was needed during debugging and a small change

PR #3054, opened 2026-02-17
feat(vscode): enhance detection for nx skills for cursor & re-add old rule file cleanup

MaxKless:


libs/vscode/mcp/src/lib/init-mcp.ts
```diff
@@ -156,6 +159,66 @@ function cleanupOldMcpJson() {
...
+      window
+        .showWarningMessage(
+          `Nx Console found outdated AI rules files (${fileNames}). Would you like to migrate to the latest agent skills?`,
+          'Migrate',
+          "Don't ask again",
+        )
+        .then((selection) => {
+          if (selection === 'Migrate') {
+            removeFiles();
+            runConfigureAiAgentsCommand();
```
juristr: Does `runConfigureAiAgentsCommand()` need to be awaited here? If it returns a promise, any errors would be silently swallowed.

libs/vscode/utils/src/lib/skill-detection.ts
```diff
@@ -1,29 +1,42 @@
...
+    return true;
+  }
+
+  return false;
+}
+
+function hasNxClaudePlugin(workspacePath: string): boolean {
+  try {
+    const settingsPath = join(workspacePath, '.claude', 'settings.json');
+    const raw = readFileSync(settingsPath, 'utf-8');
```
juristr, line 36: This does a synchronous file read + parse on every call to `hasNxWorkspaceSkill`. Could this be called frequently enough to be worth caching?

libs/nx-mcp/nx-mcp-server/src/lib/load-config-file.ts
```diff
@@ -33,7 +34,7 @@ export function loadConfigFile(workspacePath: string): NxMcpConfig {
 
   let parsed: unknown;
   try {
-    parsed = JSON.parse(raw);
+    parsed = parse(raw);
```
juristr, line 37: Nice switch to `jsonc-parser`, but the tests still only cover plain JSON. Might be worth adding a test case with comments or trailing commas to validate the JSONC behavior.

libs/vscode/utils/src/lib/skill-detection.ts
MaxKless: only happens once on startup

PR #3062, opened 2026-02-21
feat(intellij): resolve style preprocessor include paths from rspack …

gultyayev:
…config

Reads stylePreprocessorOptions.includePaths from rspack.config.ts/js and merges them with Angular project.json paths so the IDE resolves SCSS imports correctly for rspack-based projects.

Handles both default exported `createConfig` and more advanced case when `createConfig` is stored in a constant to return a modified config (e.g., when using webpack-merge).

**Before:**

image

**After:**

image

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularConfigService.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularConfigService.kt
@@ -8,6 +8,8 @@ import com.intellij.openapi.roots.ex.ProjectRootManagerEx
 import com.intellij.openapi.util.EmptyRunnable
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.vfs.VirtualFileManager
+import com.intellij.openapi.vfs.newvfs.BulkFileListener
+import com.intellij.openapi.vfs.newvfs.events.VFileEvent
 import com.intellij.util.xmlb.annotations.Attribute
 import com.intellij.util.xmlb.annotations.XMap
 import dev.nx.console.nxls.NxWorkspaceRefreshListener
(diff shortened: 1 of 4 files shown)


apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularConfigService.kt
```diff
@@ -28,11 +30,33 @@ class NxAngularConfigService(private val project: Project, private val cs: Corou
...
+                                }
+                            }
+                        }
+                    }
+                }
+            },
+        )
     }
 
+    private fun isRspackConfigFile(file: VirtualFile?): Boolean =
```
MaxKless: i think this should check against the virtual file path instead to avoid having to resolve the file for EVERY listener event.

apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularProject.kt
```diff
@@ -12,8 +12,22 @@ class NxAngularProject(
...
     private val ngProject: AngularJsonProject,
     private val workspaceFolder: VirtualFile,
+    private val projectFolder: VirtualFile? = null,
+    private val rspackConfigFile: VirtualFile? = null,
 ) : AngularProject(workspaceFolder) {
 
+    // Cached as (modificationStamp, paths) to invalidate when the file changes.
+    @Volatile private var rspackPathsCache: Pair<Long, List<String>>? = null
+
+    private fun getRspackIncludePaths(): List<String> {
```
MaxKless, line 26: can we be as defensive as possible here and wrap things in a try/catch? I want to avoid errors for users as much as possible

apps/intellij/src/main/kotlin/dev/nx/console/angular/NxAngularConfigService.kt
```diff
@@ -28,11 +30,33 @@ class NxAngularConfigService(private val project: Project, private val cs: Corou
...
+        bus.subscribe(NX_WORKSPACE_REFRESH_TOPIC, NxWorkspaceRefreshListener { refresh() })
+        bus.subscribe(
+            VirtualFileManager.VFS_CHANGES,
+            object : BulkFileListener {
+                override fun after(events: List<VFileEvent>) {
+                    if (events.any { isRspackConfigFile(it.file) }) {
+                        cs.launch {
+                            withContext(Dispatchers.EDT) {
+                                writeAction {
+                                    ProjectRootManagerEx.getInstanceEx(project)
```
MaxKless: consider deduplicating this call between here and `updateConfig` - not super important tho

MaxKless:
thanks for the fix!

gultyayev:
Thank you for the review 🙌

PR #3105, opened 2026-04-04
fix(intellij): always trigger exit handler and throttle post-crash logs

gultyayev:
When the nxls process crashes at runtime (as opposed to a startup failure), two
IntelliJ-specific issues compound the problem:

1. **`NxlsProcess.onExit` only fires when stderr has content.** If the process
   is killed by SIGKILL (OOM killer) or exits without writing to stderr, the
   callback never fires, so `NxlsWrapper.stop()` is never triggered and
   resources are not cleaned up. This was originally designed for startup error
   detection (#1649) but doesn't cover runtime crashes.

2. **"Unable to send messages to the nxls" is logged for every buffered message
   after process death** — hundreds of times from the OS pipe buffer — with no
   throttle.

## Changes

- **Always call `onExit`** when the process exits, regardless of stderr content.
  Stderr is still logged when present. Safe because `NxlsProcess.stop()` cancels
  `exitJob` in the intentional-stop path, and `NxlsWrapper.stop()` has a
  re-entrancy guard (`STOPPING` state check).
- **Add log-once guard** for the "Unable to send" message. Flag is reset on
  `start()`.

Addresses side effects of #2921. Specifically, those tons of messages that occur when the process gets killed.

## Repro of the bug

1. Pull https://github.com/gultyayev/nx-angular-mfe-repro
2. `npm i`
3. `nx sync & kill -9 <nxls pid>`

No logs when the process is killed.

After the fix

```
[L][2026-04-04 16:46:07.172] Stopping nxls
[L][2026-04-04 16:46:07.175] Unable to send messages to the nxls, the process has exited
[L][2026-04-04 16:46:07.175] Sending request to nxls: shutdown (null)
[L][2026-04-04 16:46:08.177] stopping nxls process
[L][2026-04-04 16:46:08.177] process is not alive
[L][2026-04-04 16:46:08.178] Process exited: true
```

I struggle to find "proper" reproduction for the real flood, but this small repro seems to still prove the point.

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/nxls/NxlsProcess.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/nxls/NxlsProcess.kt
@@ -46,16 +46,11 @@ class NxlsProcess(private val project: Project, private val cs: CoroutineScope) 
                 exitJob =
                     cs.launch {
                         it.awaitExit()
-                        it.errorStream.readAllBytes().decodeToString().run {
-                            if (this.isEmpty()) {
-                                return@run
-                            }
-
-                            if (project.isDisposed) {
-                                return@run
-                            }
-
-                            logger.error("Nxls early exit: $this")
+                        val stderr = it.errorStream.readAllBytes().decodeToString()
+                        if (stderr.isNotEmpty() && !project.isDisposed) {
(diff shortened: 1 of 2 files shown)


gultyayev:
The PR doesn't seem to bring much value.

PR #3129, opened 2026-05-04
fix(intellij): show actionable notification when generator filter hides all generators

### LINKED ISSUE #3128: [JetBrains 1.59.0] Generate UI silently fails: IllegalArgumentException in Notifier.notifyNoGenerators when generatorAllowlist filters out all generators
## Summary

In WebStorm, clicking **Nx Generate (UI)** does absolutely nothing — no popup, no error, no toast. The user has no way of knowing why. Tracing through plugin source revealed **two bugs**:

1. **Cover-up bug:** `Notifier.notifyNoGenerators` calls `addAction(ActionManager.getInstance().getAction("OpenLog"))`. In **WebStorm 2025.3** the action ID `"OpenLog"` does not resolve, so `getAction` returns `null`. `Notification.addAction` is `@NotNull`, so it throws `IllegalArgumentException`. The "No generators found" notification never appears, hiding the real problem from the user.
2. **Trigger bug (root cause for me):** an obsolete entry in `.idea/nx-console.xml` had a `generatorAllowlist` matching nothing in the workspace, so `getFilteredGenerators()` filtered every real generator out, producing an empty list. There is no UI affordance I could find to inspect or clear that allowlist after it gets set.

The two combined produce a perfectly silent failure on a healthy workspace.

## Environment

- **Nx Console version:** 1.59.0
- **IDE:** WebStorm 2025.3.2, Build #WS-253.30387.83
- **JDK:** 21.0.9 (JetBrains)
- **OS:** macOS 15 (Darwin 25.4.0)
- **Node:** 24.12.0 (via nvm)
- **Package manager:** pnpm
- **Nx:** 22.6.2

## Symptom

- `nx/generators` LSP request returns successfully with a full list (`@angular/cdk:drag-drop`, `@nx/angular:component`, etc. — verified in `idea.log`).
- Daemon is healthy, workspace is healthy, `nx/cloudStatus.isConnected=false` (Nx Cloud disconnected and disabled).
- Click "Generate (UI)" → nothing happens. No popup. No notification. No editor message.
- Repeated clicks produce repeated identical stack traces in `idea.log`, but no user-visible feedback.

## Stack trace (from `idea.log`)

```
SEVERE - #c.i.o.a.i.ExceptionsKt - Unhandled exception in
[Kernel@..., ..., CoroutineName(dev.nx.console.generate.NxGenerateService), Dispatchers.Default],
interactive mode: false
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'action'
of com/intellij/notification/Notification.addAction must not be null
    at com.intellij.notification.Notification.$$$reportNull$$$0(Notification.java)
    at com.intellij.notification.Notification.addAction(Notification.java)
    at dev.nx.console.utils.Notifier$Companion.notifyNoGenerators(Notifier.kt:176)
    at dev.nx.console.generate.NxGenerateService$selectGenerator$2$1.invokeSuspend(NxGenerateService.kt:73)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
    ...
SEVERE - Plugin to blame: Nx Console version: 1.59.0
SEVERE - Last Action: dev.nx.console.generate.actions.NxGenerateUiAction
```

### Bug 1 — `Notifier.kt:176`

In `apps/intellij/src/main/kotlin/dev/nx/console/utils/Notifier.kt`:

```kotlin
} else {
    notification.addAction(ActionManager.getInstance().getAction("OpenLog"))
}
```

`ActionManager.getAction("OpenLog")` returns `null` in WebStorm 2025.3 (the action ID is not registered there — it exists in IntelliJ IDEA but not in every JetBrains IDE). `Notification.addAction(@NotNull AnAction)` then throws.

**Suggested fix:** null-check the result, or guard with a `find-by-id` and fall back to no action / a different action that exists across all JetBrains IDEs (e.g., `ShowLog` or simply skip the action button when the ID is unresolved).

```kotlin
} else {
    val openLog = ActionManager.getInstance().getAction("OpenLog")
    if (openLog != null) notification.addAction(openLog)
}
```

### Bug 2 — empty `getFilteredGenerators()` with no UX path to recover

In `apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGenerateService.kt`, `getFilteredGenerators()` reads `NxConsoleProjectSettingsProvider.getInstance(project).generatorFilters` (mapped to `generatorAllowlist` in the persisted XML).

In my project, `.idea/nx-console.xml` contained:

```xml
<component name="NxConsoleProjectSettingsProvider">
  <option name="generatorAllowlist">
    <map>
      <entry key="@nx/example:*" value="true" />
    </map>
  </option>
</component>
```

I do not remember setting this and there's no `@nx/example` plugin in the workspace. The matcher uses `matchWithWildcards`, so `@nx/example:*` matches no real generator — every actual generator gets filtered out, and `selectGenerator()` calls `notifyNoGenerators(...)` (which then crashes per Bug 1).

The result is **doubly silent**: the user has no idea that (a) a filter is being applied, (b) the filter excludes everything, or (c) the resulting "no generators" notification crashed before rendering.

**Suggested fixes:**
1. Surface the active allowlist somewhere visible (settings panel, status bar tooltip, or in the empty-state notification text — "0 of N generators matched filter `@nx/example:*`").
2. Validate the allowlist against the actual generator list when settings are saved, and warn if no entry would match anything.
3. Provide a UI to clear/edit the allowlist (I could not find one in the Nx Console settings page).

## Reproduction

1. Open any Nx 22 workspace in WebStorm 2025.3 with Nx Console 1.59.0.
2. Add to `.idea/nx-console.xml`:
    ```xml
    <component name="NxConsoleProjectSettingsProvider">
      <option name="generatorAllowlist">
        <map>
          <entry key="@nx/nonexistent:*" value="true" />
        </map>
      </option>
    </component>
    ```
3. Reload the project.
4. Click any **Nx Generate (UI)** entry point (toolbar, context menu, Find Action).
5. Observe: nothing visible happens. Open `idea.log` to see the `IllegalArgumentException` from `Notifier.notifyNoGenerators`.

## Workaround that fixed it for me

Edit `.idea/nx-console.xml` to remove the allowlist:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="NxConsoleProjectSettingsProvider" />
</project>
```

Then reload from disk in WebStorm. **Generate UI works immediately** — popup appears with the full list of `@angular/cdk:*`, `@nx/angular:*`, etc. No plugin reinstall, no daemon reset, no IDE restart needed.

## Related issues

- #2753 — same user-facing symptom ("click Generate, nothing happens"), unresolved since Feb 2025; reporter never had logs.
- #2390 — adjacent symptom (Angular schematics not appearing), different stack trace, same general failure mode.

A common thread is: **when `getFilteredGenerators` returns empty, the user gets no actionable feedback in WebStorm.** Fixing Bug 1 alone would already turn the silent failure of #2753-class issues into a visible "No generators found" notification, which would have made my own diagnosis trivial.

## Asks

1. Patch `Notifier.notifyNoGenerators` (Bug 1) — small, low-risk null-guard.
2. Add UX for inspecting/clearing the generator allowlist (Bug 2) — or at least include the active filter pattern in the empty-state message.

Happy to PR Bug 1 if useful.

MaxKless:
In WebStorm 2025.3 the `OpenLog` action ID is unregistered, so the non-null `addAction` call threw and the "no generators" notification never rendered, making Nx Generate (UI) appear to do nothing when a stale `generatorAllowlist` filtered everything out. The lookup is now null-guarded, and when an active include filter hides every generator the notification names the active pattern(s) and offers a shortcut to Nx Console settings so the user can fix the allowlist.

Closes #3128

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGenerateService.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/generate/NxGenerateService.kt
@@ -70,7 +70,13 @@ class NxGenerateService(val project: Project, private val cs: CoroutineScope) {
             if (generators.isEmpty()) {
                 val hasErrors =
                     (NxlsService.getInstance(project).workspace()?.errors?.size ?: 0) > 0
-                Notifier.notifyNoGenerators(project, hasErrors)
+                val activeIncludeFilters =
+                    NxConsoleProjectSettingsProvider.getInstance(project)
+                        .generatorFilters
+                        ?.filter { it.include }
+                        ?.map { it.matcher }
+                        .orEmpty()
+                Notifier.notifyNoGenerators(project, hasErrors, activeIncludeFilters)
 
                 return@launch it.resume(null)
             }
--- a/apps/intellij/src/main/kotlin/dev/nx/console/utils/Notifier.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/utils/Notifier.kt
@@ -8,11 +8,13 @@ import com.intellij.notification.*
 import com.intellij.openapi.actionSystem.ActionManager
 import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.options.ShowSettingsUtil
 import com.intellij.openapi.project.DumbAware
 import com.intellij.openapi.project.Project
 import com.intellij.util.ui.RestartDialogImpl
 import dev.nx.console.NxConsoleBundle
 import dev.nx.console.nxls.NxRefreshWorkspaceAction
+import dev.nx.console.settings.NxConsoleSettingsConfigurable
 import dev.nx.console.telemetry.actions.TelemetryOptInAction
 import dev.nx.console.telemetry.actions.TelemetryOptOutAction
 import kotlinx.coroutines.CoroutineScope
@@ -156,24 +158,40 @@ class Notifier {
             return notification
         }
 
-        fun notifyNoGenerators(project: Project, hasNxErrors: Boolean) {
+        fun notifyNoGenerators(
+            project: Project,
+            hasNxErrors: Boolean,
+            activeFilterMatchers: List<String> = emptyList(),
+        ) {
+            val message =
+                when {
+                    activeFilterMatchers.isNotEmpty() ->
+                        "No generators matched the active generator filter (${activeFilterMatchers.joinToString(", ")}). " +
+                            "Update or remove the filter in Nx Console settings."
+                    hasNxErrors -> "No generators found. View Nx Errors for more information."
+                    else -> "No generators found. View logs for more information."
+                }
+
             val notification =
                 getGroup()
-                    .createNotification(
-                        if (hasNxErrors) "No generators found. View Nx Errors for more information."
-                        else "No generators found. View logs for more information.",
-                        NotificationType.ERROR,
-                    )
+                    .createNotification(message, NotificationType.ERROR)
                     .setTitle("Nx Console")
 
-            if (hasNxErrors) {
+            if (activeFilterMatchers.isNotEmpty()) {
+                notification.addAction(
+                    NotificationAction.createSimpleExpiring("Open Nx Console Settings") {
+                        ShowSettingsUtil.getInstance()
+                            .showSettingsDialog(project, NxConsoleSettingsConfigurable::class.java)
+                    }
+                )
+            } else if (hasNxErrors) {
                 notification.addAction(
                     NotificationAction.createSimpleExpiring("View Nx Errors") {
                         ProblemsView.getToolWindow(project)?.show()
                     }
                 )
             } else {
-                notification.addAction(ActionManager.getInstance().getAction("OpenLog"))
+                ActionManager.getInstance().getAction("OpenLog")?.let { notification.addAction(it) }
             }
             notification.notify(project)
         }
--- a/apps/intellij/src/main/kotlin/dev/nx/console/utils/Notifier.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/utils/Notifier.kt
@@ -20,6 +20,7 @@ import dev.nx.console.telemetry.actions.TelemetryOptOutAction
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import org.eclipse.lsp4j.jsonrpc.MessageIssueException
+import org.jetbrains.annotations.VisibleForTesting
 
 class Notifier {
     companion object {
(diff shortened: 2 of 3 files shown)


MaxKless:
Verified end-to-end on a real IntelliJ IDEA Ultimate `IU-253.33813.55` against a small Nx 23.2.0 fixture, and pushed regression coverage (`2af08ac7`).

**Scenario.** A Kotlin JetBrains-Driver scenario first invokes `NxGenerateUiAction` with **no** filter as a control and asserts the chooser really does list the fixture's `@nx/js` generators — so anything seen afterwards is attributable to the filter, not to an empty workspace. It then installs `generatorAllowlist = {"@nx/does-not-exist:*": true}` through the plugin's own settings state, invokes the action again, and reads the **rendered** balloon out of the live Swing hierarchy (body from `NotificationsManagerImpl$5`, action labels from `NotificationsManagerImpl$11`).

| | master `d2e4d1f7` (no fix) | master + `1d94281d` |
|---|---|---|
| notification text | `No generators found. View logs for more information.` | `No generators matched the active generator filter (@nx/does-not-exist:*). Update or remove the filter in Nx Console settings.` |
| names the active filter | **no** | **yes** |
| actions | `[Open Log in Editor]` | `[Open Nx Console Settings]` |

Both recordings show the actual Nx Console balloon in the IDE window.

**Follow-up commit.** `NoGeneratorsMessageTest` (5 tests) locks in that an active include filter is named, that every matcher is listed, that the filter message wins over the Nx-errors message, and that the two pre-existing messages are preserved verbatim when no filter is set. To assert the text without a notification bus I extracted it into `@VisibleForTesting internal fun noGeneratorsMessage`; behaviour is unchanged. `nx run intellij:test`, `ktfmtCheck` and `nx format --fix` all pass.

**One thing this does not prove.** The crash half of #3128 — `ActionManager.getAction("OpenLog")` returning `null` and blowing up `addAction` — is specific to IDEs where that action is unregistered (WebStorm 2025.3 per the reporter). On IDEA the action *is* registered, so the baseline here shows the generic balloon rather than a swallowed exception. The null-guard is reviewed by inspection only; what is demonstrated is the filter-aware message and the settings shortcut.

Videos, the verification report, and the scenario source are attached to the [Polygraph session](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/ae2f85cd-6e48-4e7c-b19f-31c6ee81dd86) (`issue-3128-repro.mp4`, `issue-3128-fixed.mp4`).

The harness is still unmerged (#3209), so it was applied as a local testing-only overlay and is not part of this branch.

MaxKless:
Heads-up on the red `Main Linux` here: it is not this branch.

The job timed out after 60 minutes with the Nx Cloud DTE execution reporting `number of completed tasks: 0` across 695 status updates — not one task finished for the whole hour, including `lint` on projects this PR doesn't touch. Agents did exist (one picked up `shared-cloud-fix-webview:build-css`), the run just never progressed. The same stall hit #3129 and a third unrelated PR in the same window, while the non-Nx-Cloud checks (CodeQL, the CodeQL Analyze jobs, `pr_title_check`) all passed.

I've re-triggered the job. If it stalls again the thing to look at is Nx Cloud for this workspace, not these commits.

MaxKless:
`Main Linux` is green on the re-run — same commit, no changes. That confirms the earlier timeout was the Nx Cloud stall rather than anything on this branch. All checks pass; this is ready for review.

leosvelperez (approved):

PR #3168, opened 2026-06-26
fix(intellij): reuse HttpClient in NxGraphServer to prevent thread/selector leak

gultyayev:
### Summary
`NxGraphServer.handleGraphRequest` created a new `java.net.http.HttpClient` on every graph request and never closed it. This change reuses a single client for the lifetime of the server and shuts it down on dispose.

### Problem
Every graph request (`requestProjectGraph`, `requestTaskGraph`, `requestExpandedTaskInputs`, `requestSourceMaps`) called `HttpClient.newBuilder().build()`. On JDK 21 an `HttpClient` is `AutoCloseable` and owns its own selector-manager thread and executor, which stay alive until the client is closed or garbage-collected. Allocating one per request leaks those threads and the associated resources, and they accumulate as the user interacts with the graph.

### Fix
- Hold a single, lazily-created `HttpClient` in the server and reuse it across requests.
- Shut it down via `shutdownNow()` in `dispose()`.
- The client lives in a lock-guarded nullable field (recreated on demand) rather than a `lazy val`, because `dispose()` is also called by `restart()`; this guarantees a restarted server transparently creates a fresh client on its next request.

### Behavior
Request semantics are unchanged — same URL, headers, async send/await, retry logic, and error handling. Only the lifecycle of the HTTP client changes.

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/graph/NxGraphServer.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/graph/NxGraphServer.kt
@@ -57,6 +57,19 @@ open class NxGraphServer(
     private var isStarted = false
     private var isStarting = false
 
+    // Reuse a single HttpClient instead of creating one per request. Each JDK
+    // HttpClient owns a selector-manager thread and an executor that live until
+    // the client is closed, so allocating one per request leaked those
+    // resources. Recreated lazily after dispose() since dispose() is also used
+    // by restart().
+    private val httpClientLock = Any()
+    private var httpClient: HttpClient? = null
+
+    private fun obtainHttpClient(): HttpClient =
+        synchronized(httpClientLock) {
+            httpClient ?: HttpClient.newBuilder().build().also { httpClient = it }
+        }
+
     init {
         with(project.messageBus.connect()) {
             subscribe(
@@ -93,7 +106,7 @@ open class NxGraphServer(
                     else -> throw Exception("unknown request type ${request.type}")
                 }
 
-            val client = HttpClient.newBuilder().build()
+            val client = obtainHttpClient()
             val httpRequest =
                 HttpRequest.newBuilder()
                     .uri(URI.create(url))
@@ -245,6 +258,10 @@ open class NxGraphServer(
         nxGraphProcess = null
         isStarted = false
         isStarting = false
+        synchronized(httpClientLock) {
+            httpClient?.shutdownNow()
+            httpClient = null
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


MaxKless:
Verified this on a real IntelliJ IDEA Ultimate `IU-253.33813.55` against a small Nx 23.2.0 fixture, and pushed one follow-up commit with regression coverage (`bdde1304`). Your fix is correct as written — I didn't change its behaviour.

**Measurement.** A Kotlin JetBrains-Driver scenario opens the Nx graph 5 times through the real `NxGraphSelectAllAction`, waits each time until the graph's JCEF document actually reports all four projects selected, and then counts live `HttpClient-<n>-SelectorManager` threads in the IDE JVM over its own JMX `ThreadMXBean`.

| after | master `d2e4d1f7` (no fix) | master + `3c51be41` |
|---|---|---|
| graph open 1 | 4 | 1 |
| graph open 2 | 4 | 1 |
| graph open 3 | 8 | 1 |
| graph open 4 | 12 | 1 |
| graph open 5 | **16** | **1** |
| closing the graph | **16** | **1** |

Baseline leaves `HttpClient-6…21-SelectorManager` alive — one per request, none released, and they survive closing the graph. With the fix a single `HttpClient-2-SelectorManager` serves all five opens.

**Follow-up commit.** `apps/intellij/src/test/kotlin/dev/nx/console/graph/NxGraphServerHttpClientTest.kt` asserts the two halves of the contract this PR introduces: repeated requests share one client, and `dispose()` terminates it so the next request after a `restart()` gets a fresh one. That needed `obtainHttpClient` widened from `private` to `@VisibleForTesting internal`; nothing else changed. `nx run intellij:test` passes, as do `ktfmtCheck` and `nx format --fix`.

Baseline/fixed videos, the full verification report, and the scenario source are attached to the [Polygraph session](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/4162990c-944d-44ec-ba3f-904b5c1a122f) (`issue-3168-repro.mp4`, `issue-3168-fixed.mp4`).

One caveat worth stating: discarded clients could in principle be reclaimed by GC, so the leak is a growth-until-collection problem rather than an unbounded one. No GC was forced in either run and the baseline count never dropped.

Verification used the plugin's registered action through Driver and JCEF inspection — no native mouse or keyboard input. The harness itself is still unmerged (#3209), so it was applied as a local testing-only overlay and is not part of this branch.

MaxKless:
Heads-up on the red `Main Linux` here: it is not this branch.

The job timed out after 60 minutes with the Nx Cloud DTE execution reporting `number of completed tasks: 0` across 695 status updates — not one task finished for the whole hour, including `lint` on projects this PR doesn't touch. Agents did exist (one picked up `shared-cloud-fix-webview:build-css`), the run just never progressed. The same stall hit #3129 and a third unrelated PR in the same window, while the non-Nx-Cloud checks (CodeQL, the CodeQL Analyze jobs, `pr_title_check`) all passed.

I've re-triggered the job. If it stalls again the thing to look at is Nx Cloud for this workspace, not these commits.

MaxKless (approved):

MaxKless:
`Main Linux` is green on the re-run — same commit, no changes. That confirms the earlier timeout was the Nx Cloud stall rather than anything on this branch. All checks pass; this is ready for review.

PR #3195, opened 2026-08-12
fix(vscode): prevent command injection via Nx Cloud CIPE branch name

bunlongheng:
## Summary

The Nx Cloud CIPE (CI Pipeline Execution) notification flow builds shell commands by string-interpolating a branch name that comes straight from the remote Nx Cloud API response. Because those commands run through `execSync` (which spawns `/bin/sh -c`), a branch name containing shell metacharacters is executed on the developer's machine.

## Details

`cipe.branch` is typed as a plain `string` (`libs/shared/types/src/lib/cloud-info.ts`) and is populated verbatim from the HTTP response in `getRecentCIPEData`:

```ts
const responseData = JSON.parse(response.responseText) as {
  ciPipelineExecutions: CIPEInfo[];
  workspaceUrl: string;
};
```

The value is never validated against git ref-name rules. It then reaches three shell commands:

- `libs/vscode/nx-cloud-view/src/cipe-notification-service.ts`
  ```ts
  execSync(`git rev-parse --verify origin/${targetBranch}`, { cwd: getNxWorkspacePath() });
  ```
- `libs/vscode/nx-cloud-view/src/nx-cloud-fix-webview.ts`
  ```ts
  execSync(`git pull --ff-only origin ${targetBranch}`, { cwd });
  execSync(`git fetch origin ${targetBranch}:${targetBranch}`, { cwd });
  ```

The `rev-parse` call is the most severe: it runs automatically inside `showAppliedFixNotification`, which is reached from `processCIPENotifications` during background CIPE polling. No user interaction is required. As soon as the extension observes an applied AI fix for a branch, the branch name is passed through the shell.

Git ref names permit `$`, `` ` ``, `(`, `)`, `;`, `&`, `|`, `{`, `}` and so on, so a branch like `$(touch${IFS}/tmp/pwned)` is both a valid branch and a working command-injection payload. An attacker who can get such a branch into a workspace connected to Nx Cloud (for example a fork PR that triggers a self-healing CI run) can achieve arbitrary command execution on any collaborator whose editor polls that CIPE data.

## Fix

Replace the three interpolated `execSync` calls with `execFileSync('git', [...args])`. `execFileSync` does not spawn a shell, so the branch is passed to `git` as a single literal argument and metacharacters carry no special meaning. Legitimate branch names behave exactly as before; a malformed ref simply fails the git command as it did previously.

No API or test changes are required.

### DIFF
--- a/libs/vscode/nx-cloud-view/src/cipe-notification-service.ts
+++ b/libs/vscode/nx-cloud-view/src/cipe-notification-service.ts
@@ -7,7 +7,7 @@ import {
 import { getTelemetry } from '@nx-console/vscode-telemetry';
 import { commands, window } from 'vscode';
 import { fetchAndPullChanges } from './nx-cloud-fix-webview';
-import { execSync } from 'child_process';
+import { execFileSync } from 'child_process';
 
 export class CIPENotificationService {
   private sentNotifications = new Set<string>();
@@ -242,7 +242,7 @@ export class CIPENotificationService {
     const targetBranch = cipe.branch;
     let hasBranchOnRemote: boolean;
     try {
-      execSync(`git rev-parse --verify origin/${targetBranch}`, {
+      execFileSync('git', ['rev-parse', '--verify', `origin/${targetBranch}`], {
         cwd: getNxWorkspacePath(),
       });
       hasBranchOnRemote = true;
--- a/libs/vscode/nx-cloud-view/src/cipe-notifications-service.spec.ts
+++ b/libs/vscode/nx-cloud-view/src/cipe-notifications-service.spec.ts
@@ -1,5 +1,7 @@
 import { CIPEInfo } from '@nx-console/shared-types';
-import { window } from 'vscode';
+import { WorkspaceConfigurationStore } from '@nx-console/vscode-configuration';
+import { execFileSync, execSync } from 'child_process';
+import { ExtensionContext, window } from 'vscode';
 import { CIPENotificationService } from './cipe-notification-service';
 
 const globalConfigMock = jest.fn().mockReturnValue('all');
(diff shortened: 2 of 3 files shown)


MaxKless:
Reproduced the injection in a real editor and verified this fix closes it, then pushed one regression-test commit onto this branch (`c0a23a08`, on top of your `e7474285` — ordinary fast-forward, nothing rewritten).

**Reproduction**, real VS Code stable 1.137.0 (Playwright Electron) with the built extension, pointed at a fixture whose `nxCloudUrl` is a local mock Nx Cloud. The mock returns a failed CIPE with an automatically-applied AI fix on the branch `fix/cipe$(touch ${IFS}<marker>)`. That path (`processCIPENotifications` → `showAppliedFixNotification`) runs with no user interaction.

| Nx Console | shell `git rev-parse … origin/${branch}` in the built `main.js` | notification | marker file created |
| --- | --- | --- | --- |
| master `f1425880` | present | shown | **YES** — the branch name executed in a shell |
| `f1425880` + this PR | gone | shown (identical text) | **NO** |

The notification text is the same in both runs; only the side effect differs. Videos, screenshots, the mock, the launcher, and the fixture are attached to the maintenance session below.

**Regression test** (`c0a23a08`): asserts the applied-fix path calls `execFileSync('git', ['rev-parse', '--verify', 'origin/' + branch], { cwd })` and never `execSync` for a `$(touch …)` branch, and that the pull action is only offered when the branch exists. 67 passed with the fix; the two new tests fail when the fix is reverted (2 failed, 65 passed). `nx format:check` is clean.

CI on this fork PR is still `action_required` (awaiting a maintainer to approve the workflow run), so the suite has not run here yet.

Evidence: https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/gentle-penguin-2a29bad3

MaxKless (approved):

PR #3209, opened 2026-09-07
fix(intellij): preserve saved run arguments and add recorded automation

MaxKless:
Launching an Nx target from the JetBrains tree currently erases saved CLI arguments. Preserve those arguments when reusing a configuration without an explicit command; Run Anything can still replace or clear them. Fixes #3191.

Add a separate Kotlin Driver source set and sandbox launcher so local agents can inspect the IDE, invoke real plugin actions, and verify changes. Record the IDE window to MP4 even when assertions fail, with reusable graph and run-argument scenarios and setup instructions in `apps/intellij/AUTOMATION.md`. Automation sources are excluded from the shipped plugin.

Include Kotlin sources and compiled dependency classes explicitly in the `instrumentCode` cache inputs. This prevents stale plugin bytecode after a rebuild while keeping the build step cacheable for distributed CI on Nx Agents.

Validation:
- Regression baseline: 6 focused tests, 1 failure. Fixed: all 6 pass, including explicit argument replacement/clearing. Clean up saved configurations between tests.
- Real IntelliJ IDEA 2025.2.2 run: the tree Run action previously erased the saved greeting and the Node task received `[]`; after rebuilding, the setting survives and the task receives `["--greeting=hello from Nx Console"]`.
- Captured a 9.25-second repro video and a 9.08-second fixed video using the same fixture, tree action, and behavioral assertions.
- Nx formatting, Kotlin formatting/checks, and `git diff --check` pass.

The graph scenario documents a separate upstream Nx focus-to-full-graph regression; this PR does not fix that Nx graph state bug. GUI verification used the tree's registered Run action through Driver, not native double-click input in PhpStorm.

Video evidence is attached to the [automatically created Polygraph session](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/d48e78ac-76ee-469f-9f48-0b30f45f33fa):
- `issue-3191-repro.mp4` — failing baseline (9.25 seconds): saved arguments erased; target receives `[]`.
- `issue-3191-fixed.mp4` — passing verification (9.08 seconds): saved arguments preserved and received by the target.

PR #3211, opened 2026-09-08
fix(nxls): describe and validate the whole of project.json

### LINKED ISSUE #1986: JSON schema when open project.json error in Webstorm
## Steps to Reproduce

```sh
npx create-nx-workspace@latest acme-corp
```
1. npp 
2. `npx nx g @nx/js:library --name=sample --unitTestRunner=none --directory=libs/sample --projectNameAndRootFormat=as-provided --no-interactive`

### Environment

- Nx Console version:
- VSCode version:

MaxKless:
Closes #1986

## The bug

Completion for `project.json` in JetBrains IDEs comes from nxls, and the schema nxls serves described only the five properties that need workspace knowledge:

```ts
// libs/shared/json-schema/src/lib/project-json-schema.ts
properties: { sourceRoot, implicitDependencies, tags, namedInputs, targets }
```

Everything else Nx supports was invisible — no completion and no hover documentation for `name`, `projectType`, `generators`, `metadata`, `release` or `root`. That is @Den-dp's report. It affects VS Code equally, since it goes through the same server.

On top of that, nxls never validated anything, so a `project.json` could contradict Nx's schema without any editor saying so, and the IntelliJ client's `publishDiagnostics` was `TODO("Not yet implemented")` — it would have thrown had the server ever sent one.

## The fix

**nxls** merges Nx's own `schemas/project-schema.json` — read from the installed package, so it tracks the Nx version in use — into the schema it serves, with the workspace-aware entries winning where the two overlap. `getPackageJsonSchema` already did exactly this for the `nx` key inside `package.json`; this reuses the pattern for `project.json` itself, so both files now describe the same thing.

It also runs the json language service's validation for `project.json` and `nx.json` and publishes the result. Problems are **warnings**, not errors: the schema is assembled from the installed package plus whatever executor schemas the workspace resolves, so a mismatch is a strong hint rather than a certainty.

**The IntelliJ client** stores what arrives and renders it through an `ExternalAnnotator`, which is what makes the validation visible in the editor.

## Deliberately out of scope

The IDE's own schema mapping is untouched. JetBrains IDEs still resolve SchemaStore's `project.json` entry for this file — that entry is for .NET Core's unrelated format, but displacing a SchemaStore mapping is a decision about every `project.json` on the machine, not just Nx's, so this change leaves it alone and adds the Nx knowledge through the server instead.

## Verification

Real IntelliJ IDEA `IU-253.33813.55`, same fixture and same automation scenario for both runs. Before/after table, videos and raw results in the comment below.

---

### DIFF
--- a/apps/intellij/src/main/kotlin/dev/nx/console/nxls/NxlsDiagnosticsAnnotator.kt
+++ b/apps/intellij/src/main/kotlin/dev/nx/console/nxls/NxlsDiagnosticsAnnotator.kt
@@ -1,0 +1,72 @@
+package dev.nx.console.nxls
+
+import com.intellij.lang.annotation.AnnotationHolder
+import com.intellij.lang.annotation.ExternalAnnotator
+import com.intellij.lang.annotation.HighlightSeverity
+import com.intellij.openapi.editor.Document
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiDocumentManager
+import com.intellij.psi.PsiFile
+import dev.nx.console.utils.DocumentUtils
+import org.eclipse.lsp4j.Diagnostic
+import org.eclipse.lsp4j.DiagnosticSeverity
+import org.eclipse.lsp4j.Position
+
+class NxlsDiagnosticsAnnotator : ExternalAnnotator<List<Diagnostic>, List<Diagnostic>>() {
+
+    override fun collectInformation(file: PsiFile): List<Diagnostic>? {
+        if (!DocumentUtils.isNxFile(file.name)) {
+            return null
+        }
+        val virtualFile = file.virtualFile ?: return null
+        return NxlsDiagnosticsService.getInstance(file.project)
+            .diagnosticsFor(virtualFile)
+            .ifEmpty { null }
+    }
+
+    override fun doAnnotate(collectedInfo: List<Diagnostic>?): List<Diagnostic>? = collectedInfo
+
+    override fun apply(
+        file: PsiFile,
+        annotationResult: List<Diagnostic>?,
+        holder: AnnotationHolder,
+    ) {
+        if (annotationResult.isNullOrEmpty()) {
+            return
+        }
+        val document = PsiDocumentManager.getInstance(file.project).getDocument(file) ?: return
+        annotationResult.forEach { diagnostic ->
+            val range = diagnostic.range.toTextRange(document) ?: return@forEach
+            holder
+                .newAnnotation(severityOf(diagnostic.severity), diagnostic.message)
+                .range(range)
+                .create()
+        }
+    }
+
+    private fun severityOf(severity: DiagnosticSeverity?): HighlightSeverity =
+        when (severity) {
+            DiagnosticSeverity.Error -> HighlightSeverity.ERROR
+            DiagnosticSeverity.Information,
+            DiagnosticSeverity.Hint -> HighlightSeverity.INFORMATION
+            else -> HighlightSeverity.WARNING
+        }
+}
+
+private fun org.eclipse.lsp4j.Range.toTextRange(document: Document): TextRange? {
+    val start = position(document, start)
+    val end = position(document, end)
+    // A zero-width range would not be visible, so fall back to highlighting to the end of the line.
+    if (end > start) {
+        return TextRange(start, end)
+    }
+    val lineEnd = document.getLineEndOffset(document.getLineNumber(start))
+    return if (lineEnd > start) TextRange(start, lineEnd) else null
+}
+
+private fun position(document: Document, position: Position): Int {
+    val line = position.line.coerceIn(0, (document.lineCount - 1).coerceAtLeast(0))
+    val lineStart = document.getLineStartOffset(line)
+    val lineEnd = document.getLineEndOffset(line)
+    return (lineStart + position.character).coerceIn(lineStart, lineEnd)
+}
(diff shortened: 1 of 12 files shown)


MaxKless:
## Before / after

Real IntelliJ IDEA `IU-253.33813.55` on a fixture workspace with nx 23.2.0 installed. The scenario file is byte-identical for both runs — the fixed run was recorded first, then the plugin change was reverted, the IDE rebuilt and relaunched, and the baseline re-recorded with the same scenario. Baseline was checked to contain no `dev/nx/console/ide/json_schema` classes and no `JavaScript.JsonSchema` extension in the deployed plugin jar.

It creates `<label>-project/project.json` inside the workspace, opens it, and reads three things through the JetBrains Driver: which schema `JsonSchemaService` resolves, what the real `CodeCompletion` action puts in the lookup with the caret on an empty line at the top level, and what `DaemonCodeAnalyzer` reports for a deliberately out-of-enum `"projectType": "libraryyy"`.

| | `master` | this branch |
|---|---|---|
| resolved schema | `www.schemastore.org/project.json` | `<workspace>/node_modules/nx/schemas/project-schema.json` |
| completion items | **51** | **8** |
| what is offered | `authors, buildOptions, compilationOptions, frameworks, iconUrl, licenseUrl, packInclude, packOptions, publishOptions, requireLicenseAcceptance, runtimes, testRunner, userSecretsId, version, …` | `generators, metadata, namedInputs, release, root, tags` |
| `"projectType": "libraryyy"` | not reported | `Value should be one of: "library", "application"` |
| scenario | FAIL | PASS |

The baseline completion list is the smoking gun — those are .NET Core project properties, exactly the schema @Den-dp identified in the issue.

Being precise about two things:

- `namedInputs` and `tags` show up on both sides (`tags` exists in the .NET schema too, and Nx Console's own nxls-backed completion contributor adds a couple of templated entries), so they don't discriminate on their own. `generators` and `release` do, as do the resolved schema file and the validation result.
- The recorder samples the main IDE window at ~2–4 fps and does not capture separate popup windows, so **the completion popup itself is not in the video** — those numbers come from reading the live `LookupManager` through the driver. What the video does show as native product UI is the JSON schema widget in the status bar (`JSON: project.json` → `JSON: Nx project`), the inspection widget going from a lone spell-check hit to ⚠ 1, and `"libraryyy"` picking up an underline and a gutter stripe.

Videos, raw scenario output, the assertion failure and the scenario source (with the testing-only harness overlay from the still-unmerged #3209, so this is reproducible before that lands) are attached to the Polygraph session:
https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/b5ecba22-02fd-4f79-bf60-bcb89b1cf377

## Tests

`./gradlew :intellij:test` is green, including five new `NxJsonSchemaProviderFactoryTest` cases: serves the installed schema for workspace files, ignores `project.json` under `node_modules`, ignores `project.json` outside the workspace, ignores other json files, and reports no schema file when Nx is not installed.

---

MaxKless:
## Reworked: the fix now lives in nxls

The first revision of this PR registered an IntelliJ `JsonSchemaFileProvider` that displaced SchemaStore's `project.json` mapping. That was the wrong layer — completion for `project.json` already comes from nxls, and displacing a SchemaStore mapping is a decision about every `project.json` on the machine, not just Nx's. Reworked so the IDE's schema mapping is untouched and the Nx knowledge arrives through the server.

## Before / after

Real IntelliJ IDEA `IU-253.33813.55`, nx 23.2.0 in the fixture. Byte-identical scenario for both runs; the fixed run was recorded first, then the change reverted, the IDE rebuilt and relaunched, and the baseline re-recorded. The baseline build was checked to contain no `NxlsDiagnosticsAnnotator`, no `externalAnnotator` registration and neither nxls change.

| | `master` | this branch |
|---|---|---|
| schema the IDE resolves | `www.schemastore.org/project.json` | **unchanged** |
| completion items at the top level | 51 | 55 |
| properties only Nx's schema describes | none | `generators`, `metadata`, `release`, `root` |
| `"projectType": "libraryyy"` | not reported | `Value is not accepted. Valid values: "library", "application".` |
| scenario | FAIL | PASS |

Two things worth being precise about:

- The baseline's 51 items are SchemaStore's .NET Core properties (`authors`, `frameworks`, `iconUrl`, `packInclude`, `requireLicenseAcceptance`, `testRunner`, `userSecretsId`, …) plus `namedInputs` and `tags`, which nxls already knew. Those two are not discriminating, which is why the assertion uses `generators`/`metadata`/`release`/`root`.
- The reported message is the vscode-json-languageservice wording, not IntelliJ's own (`Value should be one of: …`). That is what proves it arrived over LSP through the new annotator rather than from the IDE's own JSON machinery.

The recorder samples the main IDE window at ~2–4 fps and does not capture popup windows, so **the completion popup is not in the video** — those counts were read from the live `LookupManager` through the driver. What the video shows as native UI is the inspection widget going from a lone spell-check hit to a warning, and `"libraryyy"` picking up an underline and a gutter stripe, with the status bar still reading `JSON: project.json` in both.

Videos, raw scenario output and the scenario source are on the Polygraph session:
https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/b5ecba22-02fd-4f79-bf60-bcb89b1cf377

## Tests

`nx run-many -t lint,typecheck,test -p nxls,shared-json-schema,language-server-capabilities-code-completion` and `./gradlew :intellij:test` are green, including five new `merge-schemas.spec.ts` cases: keeps the properties only the installed schema describes, lets the workspace-aware schema win on overlap, keeps `definitions` so internal `$ref`s still resolve, drops the installed schema's `$schema`/`$id`, and falls back when Nx is not installed.

## Note for whoever picks this up

Validation is new behaviour on a file people already have, and the merged schema pulls in executor schemas via `$ref`. I set `schemaValidation: 'warning'` for that reason, and scoped it to `project.json` and `nx.json`. If a workspace with unusual executor schemas starts showing noise, that setting is the dial.

---

MaxKless:
Rebased onto `97796b87` (#3215), which removes the `nxls-e2e` fixture-install failure that was the only thing red here. Clean replay of both commits; the diff against master is unchanged.

PR #3212, opened 2026-09-09
chore(repo): migrate to nx 23.3.0-beta.0

FrozenPandaz:
## Current Behavior

The repo is on nx `23.2.0-rc.2`.

## Expected Behavior

The repo is on nx `23.3.0-beta.0`. `nx` and all 16 `@nx/*` packages are bumped to the exact target version. The separately-versioned `21.0.0-beta.0` powerpack packages are intentionally untouched.

`nx migrate` reported no migrations to run for this hop, so this is a dependency-only change (`package.json` + `yarn.lock`).

Verified locally with `--skip-nx-cache` (project graph fully resolved):
- `lint`: 59 projects pass
- `typecheck`: 64 projects + 1 dep pass
- `build` (`--exclude=intellij`): 11 projects + 48 deps pass

## Related Issue(s)

N/A

---
<p><a href="https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/Migrate-repos-to-nx-23.3.0-beta.0-4968a04a">View Polygraph session ↗</a></p>


AgentEnder (approved):

MaxKless:
Rebased onto `97796b87` so this picks up #3215. Clean replay of all six commits, no conflicts, and I checked that the net effect on `libs/shared/e2e-utils/src/lib/utils.ts` is empty — the `install e2e workspaces with legacy peer deps` / `Revert` pair cancels out and master's `npm_config_legacy_peer_deps` line survives intact.

One thing worth flagging on `chore(repo): bump nx cloud agent image to node 20.19`, because it reads as if it replaces that line:

> `ubuntu22.04-node20.19-v3` ships node 20.19.1 / npm 10.8.2, which resolves the tree.

I can't reproduce that. On npm 10.8.2, with `npm_config_legacy_peer_deps` explicitly unset, the fixture command still dies:

```
$ npm --version
10.8.2
$ CI=true npx create-nx-workspace@22.7.0 ws --preset=react-standalone --bundler=vite \
    --e2eTestRunner=cypress --style=css --nxCloud=skip --pm=npm --no-interactive

Failed to create a workspace: npm error Cannot read properties of null (reading 'edgesOut')
```

and it is the same code path, not a different one:

```
silly fetch manifest vite@^6.0.0 || ^7.0.0 || ^8.0.0
verbose stack TypeError: Cannot read properties of null (reading 'edgesOut')
    at #loadPeerSet (@npmcli/arborist/lib/arborist/build-ideal-tree.js:1286:38)
```

I bisected this while working on #3215: 10.2.4, 10.8.2, 10.9.4, 11.0.0 and 11.5.2 all crash; only 11.19.0 installs it without `legacy-peer-deps`. So the image bump does not look sufficient on its own for the `react-standalone` fixture that both `nxls-e2e`'s `generator-options` and `nx-mcp-e2e`'s `modern-nx-layout` use.

That may well differ from what you hit — your commit describes the crash in the nx-mcp suite and diagnoses it as `@nx/devkit`'s `nx` peer rather than vitest's optional peers, and the resolved tree has moved since. Not suggesting the bump should come out: node 20.11 → 20.19 is worth having regardless. Just that the `utils.ts` line is still load-bearing, so it shouldn't be dropped later on the assumption the image fixed it.

Nothing to change here as far as I can tell — flagging it so the commit message doesn't mislead whoever reads it next.

MaxKless:
Small correction to my comment above, same as the one I've left on #3215: I described the agents as shipping **npm 10.2.4**, inferred from the node 20.11 image. The EBADENGINE warnings in the runs on this branch show it is actually **npm 8.19.4** on node 20.11.1.

Doesn't affect the point — 8.19.4 is below the whole range I bisected, and 10.2.4 through 11.5.2 all crash on this tree, so the `utils.ts` line is still doing the work — but the version number I attached to the agents was an inference, not a measurement, and it was wrong.

MaxKless (approved):

Issue #3193, opened 2026-08-10
Nested projects not shown in Projects tree view; visible in list view and PDV (regression)


vttta:
## Current Behavior

In the Nx Console **PROJECTS** panel using the **tree** viewing style (`nxConsole.projectViewingStyle: "automatic"`, which resolves to tree), projects whose root directory is **nested under another project's root** are not displayed at all.

Our workspace has aggregator projects at `e2es/parent-*`, e.g. `e2es/parent-d` (root = `e2es/parent-d`) and many child projects nested beneath it, e.g. `e2es/parent-d/child-a-e2e`. The tree renders only the parent (`parent-d-e2e`) as a leaf and hides every nested child. The children (and their `e2e` run targets) are therefore unreachable from the tree.

<img width="503" height="296" alt="Image" src="https://github.com/user-attachments/assets/861bd96b-ddde-4624-8d67-7e5f2c4bc60c" />

This is purely a tree-rendering problem — the project graph is correct:
- The **Project Details View** for a hidden child (e.g. `parent-d-child-a-e2e`) shows all its targets.
<img width="460" height="720" alt="Image" src="https://github.com/user-attachments/assets/013636b0-3ec1-4b11-85c7-9835c309d750" />

- `nx show project parent-d-child-a-e2e` lists it with fully-resolved targets.
- Switching `nxConsole.projectViewingStyle` to `list` makes **every** nested child appear individually in the panel.

## Expected Behavior

Nested projects should be visible in the tree view — the parent should be shown both as a project and as an expandable node containing its nested child projects (or the children should otherwise be reachable in the tree).

**Is this a regression?** Yes. Nested projects rendered correctly in the tree up to and including **18.24.1**, which was the last version where this worked. We pinned to 18.24.1 for a long time specifically to keep this working, but eventually an unrelated problem made that old version unusable, forcing us to upgrade — and on the current version the nested projects no longer show in the tree. Current broken version is **18.101.1**.

## Steps to Reproduce

**Reproduction Repository/Steps:**

1. In an Nx workspace, create a project at `e2es/parent` (project root = `e2es/parent`).
2. Create child projects nested beneath it: `e2es/parent/child-a-e2e`, `e2es/parent/child-b-e2e`, each with its own project.json.
3. Ensure `nxConsole.projectViewingStyle` is `automatic` or `tree`.
4. Open the Nx Console **PROJECTS** panel and expand the e2es folder.

[...]

PR #3216, opened 2026-09-10
fix(nxls): keep a top-level project a project when other projects nest inside it

MaxKless:
## Current Behavior

In the tree layout of the projects view, a project whose root is a **top-level directory that also contains other projects** is rendered as a plain folder. Its targets, run/debug actions, "Show project configuration" and project details are all unreachable from the tree, even though the project is in the graph and the projects nested below it still render fine.

Concretely, for a workspace with no project at `.` and:

```
packages/project.json      → packages-aggregator   (target: hello)
packages/a/project.json    → child-a
packages/b/project.json    → child-b
```

the tree shows `packages` as a folder, and `hello` is simply not there:

```
Projects / packages                  ← folder, not the project
Projects / packages / child-a
Projects / packages / child-a / hello
Projects / packages / child-b
Projects / packages / child-b / hello
```

## Cause

`getProjectFolderTree` kept its roots in a `Set<TreeNode>`, holding node *objects*:

```ts
const roots = new Set<TreeNode>();
…
if (projectConfiguration && projectName && treeMap.has(dir)) {
  const oldNode = treeMap.get(dir)!;
  treeMap.set(dir, { dir, projectName, projectConfiguration, children: oldNode.children });
  return;                       // ← treeMap now holds a new object; roots still holds the old one
}
```

When `packages/a` is processed first, `packages` is created as a *folder* node and — being a single-segment directory with no parent — is added to `roots`. When `packages` itself is processed, the map entry is replaced by a project node, but the set still holds the discarded folder node.

The children array survives (it is passed by reference into the new node), which is why the nested projects still appear. What is lost is the identity: `roots[0]` comes back with `projectName: undefined` and `projectConfiguration: undefined`.

Both tool windows decide between a folder and a project purely from that field — IntelliJ in `NxFolderTreeBuilder.createFolderOrProjectNode`, VS Code in `TreeView.createFolderOrProjectTreeItemFromNode` — so both render the folder.

Two conditions have to hold, which is why this is not more visible: the directory has to be **single-segment** (only those reach `roots` directly — `e2es/parent` is attached to its `e2es` parent instead), and there has to be **no project at `.`** (the workspace-root special case rebuilds the root list from the map and hides it).

Verified against nxls directly, before the fix:

```
roots:   [{ dir: "packages", projectName: null, hasConfig: false, children: ["packages/a","packages/b"] }]
treeMap: [ …, { dir: "packages", projectName: "packages-aggregator", hasConfig: true, … } ]
```

## Expected Behavior

A root is the same node the tree map holds for that directory.

Roots are now tracked by directory and resolved against the map at the end, which cannot go stale. The workspace-root special case becomes a `rootDir === '.'` comparison instead of reaching into `projectConfiguration.data.root`.

The tree building moved into `build-project-folder-tree.ts` so it can be exercised without a workspace; `getProjectFolderTree` is now just the `nxWorkspace` call plus that function.

## Verification

Real IntelliJ IDEA `IU-253.33813.55`, fixture with nx 23.2.0 and 13 projects (above the ten-project threshold where the automatic tool window style switches from the flat list to the folder tree), no project at the workspace root. Same scenario file for both runs — the fixed run was recorded first, then the two product files were reverted to `master`, nxls rebuilt, the IDE relaunched, and the baseline recorded. The rebuilt `nxls/main.js` timestamp was checked against the source edit each time.

`ReproFolderTreeRootsKt` expands the whole tree, scrolls the node under test into view, and asserts on the paths the tree actually renders.

| | `master` (`5b71ed29`) | this branch (recorded at `f168c858`, rebased to `dfd97b68`) |
| --- | --- | --- |
| top-level node | `packages` (folder icon) | `packages-aggregator` (project icon) |
| its children | `child-a`, `child-b` | `hello`, `child-a`, `child-b` |
| `hello` target reachable | **no** | yes |
| scenario | FAIL | PASS |

```
The project rooted at 'packages' is rendered as the folder 'packages'.
Top-level nodes: [libs, packages]
```

Videos, both `result.txt` files, the full scenario output and the fixture description are attached to the Polygraph session:
https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/able-tapir-a8aaa1b5

The recorded frames show the selected row directly: the folder `packages` followed immediately by `child-a` before the fix, and `packages-aggregator` followed by its `hello` target after it. Verification is through the driver reading the live `JTree`'s expanded paths plus the recorded frames — no native mouse or keyboard input.

## Tests

`libs/language-server/workspace` had no tests; this adds its first spec, seven cases over `buildProjectFolderTree`, running in ~0.2s. Three of them fail against the previous implementation:

- keeps a top-level project a project when a nested project is added first
- does not depend on the order projects arrive in
- returns the same node object for a root as for its treeMap entry

`nx run-many -t test,lint,typecheck -p language-server-workspace` passes. `nx format --fix` and `nx run-many -t ktfmtFormat` applied.

## Notes for reviewers

- The second commit is separable and is test tooling only. `recordIde` could not capture a single frame on IntelliJ 2025.3 — the IDE writes each capture into `<index>_<name>` under the log's `screenshots` directory while the recorder resolved the bare name, so every recording died on frame 0 with `No main IDE window captured`. Without it there is no video evidence for this or any other fix on this IDE build.
- The existing `project-folder-tree` e2e cannot catch this: its workspace is created from a preset that has a project at `.`, and the workspace-root special case masks the stale root. That is why the regression coverage is a unit test rather than another e2e case.
- The fix is in shared nxls code, so VS Code's tree gets it too. I verified the defect by reading `TreeView.createFolderOrProjectTreeItemFromNode`, but only reproduced and recorded it in IntelliJ — the Kotlin harness does not drive VS Code.
- Rebased onto `97796b87` now that #3215 has landed, so the `nxls-e2e` fixture-install failure that was blocking every nxls-touching PR is gone. The rebase was a clean replay of both commits; the diff against `master` is byte-identical to what was recorded.

## Related Issue(s)

N/A — found while investigating #3193. That report is a different case (`e2es/parent` is two segments deep, so it never reaches the roots set) and both tree renderers already handle it; this is the neighbouring case that is still broken.

---
<p><a href="https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/able-tapir-a8aaa1b5">View Polygraph session ↗</a></p>

### DIFF
--- a/apps/intellij/AUTOMATION.md
+++ b/apps/intellij/AUTOMATION.md
@@ -217,6 +217,32 @@ The tree Run action uses the SDK's EDT/read-action context, matching
 `Driver.invokeAction`. Calling `actionPerformed` directly on the EDT can fail
 IntelliJ's write-intent lock checks during execution startup.
 
+### Folder tree roots
+
+`ReproFolderTreeRootsKt` checks that a project whose root is a top-level
+directory still renders as a project when other projects are nested inside it.
+It expands the whole Nx Console tree, scrolls the node under test into view, and
+asserts on the paths the tree actually renders.
+
+Use a fixture with **more than ten projects**, so the automatic tool window style
+resolves to the folder tree rather than the flat list, and with no project at the
+workspace root. It needs a project rooted at `packages` named
+`packages-aggregator` with a `hello` target, two projects nested below it at
+`packages/a` and `packages/b` (`child-a`, `child-b`), and enough other projects
+elsewhere to clear the ten-project threshold.
+
+```sh
+NX_AUTOMATION_LABEL=issue-folder-tree-roots-repro CI=true \
+  JAVA_TOOL_OPTIONS='-XX:ActiveProcessorCount=4 -Dorg.gradle.workers.max=2 -Dorg.gradle.priority=low' \
+  yarn nx run intellij:runAutomation --batch=false --parallel=2 \
+  --args='--max-workers=2 --priority=low -PautomationMain=dev.nx.console.automation.ReproFolderTreeRootsKt'
+```
+
+Rerun with `NX_AUTOMATION_LABEL=issue-folder-tree-roots-fixed` after rebuilding
+and restarting the IDE. The video shows the selected tree row: the folder
+`packages` with no target before the fix, and the project `packages-aggregator`
+with its `hello` target and both nested projects after it.
+
 References:
 
 - [JetBrains Driver SDK](https://github.com/JetBrains/intellij-community/blob/master/tools/intellij.tools.ide.starter.driver/README.md)
--- a/apps/intellij/src/automation/kotlin/dev/nx/console/automation/RecordIde.kt
+++ b/apps/intellij/src/automation/kotlin/dev/nx/console/automation/RecordIde.kt
@@ -3,6 +3,7 @@ package dev.nx.console.automation
 import com.intellij.driver.client.Driver
 import com.intellij.driver.client.utility
 import com.intellij.driver.sdk.jdk.getSystemProperty
+import java.nio.file.Files
 import java.nio.file.Path
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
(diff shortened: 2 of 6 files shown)


MaxKless:
CI on `f168c858`: one failed task, and it is the repo-wide e2e blocker rather than anything here.

```
✔ Completed   language-server-workspace:test    2s   Cache Miss     ← the new tests
✖ Failed      nxls-e2e:e2e-ci--src/generators/gener…   23s   Cache Miss
```

That spec dies in `beforeAll` with `npm ERR! Cannot read properties of null (reading 'edgesOut')` while installing its fixture workspace — the only occurrence of `edgesOut` in the run. It fails on every PR that touches nxls; #3215 fixes it. Once that lands, a rerun here should be green.

Everything else on this run passed, including the seven new `buildProjectFolderTree` cases.

leosvelperez (approved):

